PR #67, opened 2026-01-18
Feat/cashu ts v3

TheRealCheebs:
This is the start of the migration to cashu-ts v3.

It is still a work in progress but as its larger it would be good to have this opened for feedback / notes / following along as the changes proceed.

### DIFF
--- a/packages/adapter-tests/src/integration.ts
+++ b/packages/adapter-tests/src/integration.ts
@@ -9,12 +9,13 @@ import {
   PaymentRequestTransportType,
   type MintKeys,
   type Token,
+  type HasKeysetKeys,
 } from '@cashu/cashu-ts';
 import { createFakeInvoice } from 'fake-bolt11';
 
 export type OutputDataFactory = (
   amount: number,
-  keys: MintKeys | { id: string; unit?: string },
+  keys: MintKeys | HasKeysetKeys,
 ) => OutputData;
 
 export type IntegrationTestRunner = {
--- a/packages/core/models/History.ts
+++ b/packages/core/models/History.ts
@@ -13,14 +13,14 @@ export type MintHistoryEntry = BaseHistoryEntry & {
   paymentRequest: string;
   quoteId: string;
   state: MintQuoteState;
-  amount: number | null;
+  amount: number;
 };
 
 export type MeltHistoryEntry = BaseHistoryEntry & {
   type: 'melt';
   quoteId: string;
   state: MeltQuoteState;
-  amount: number | null;
+  amount: number;
 };
 
 /**
--- a/packages/core/services/HistoryService.ts
+++ b/packages/core/services/HistoryService.ts
@@ -247,7 +247,7 @@ export class HistoryService {
       quoteId,
       state: quote.state,
       createdAt: Date.now(),
-      amount: quote.amount ?? null,
+      amount: quote.amount,
     };
     try {
       await this.historyRepository.addHistoryEntry(entry);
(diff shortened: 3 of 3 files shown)


packages/core/services/MeltQuoteService.ts
```diff
@@ -149,9 +149,14 @@ export class MeltQuoteService {
...
         );
-        const { send, keep } = await wallet.send(outputData.sendAmount, selectedProofs, {
-          outputData,
-        });
+        const outputConfig: OutputConfig = {
+          send: { type: 'custom', data: outputData.send },
+          keep: { type: 'custom', data: outputData.keep },
+        };
+
+        // Note: The custom OutputType does not support automatic fee inclusion, it shouldn't be added in the
```
Egge21M: Can you add additional context? Fee inclusion is done by the output creation in proof service, no?

packages/core/services/WalletService.ts
```diff
@@ -133,6 +143,7 @@ export class WalletService {
         this.logger && this.logger.child ? this.logger.child({ module: 'Wallet' }) : undefined,
       bip39seed: seed,
     });
+    await wallet.loadMint();
```
Egge21M: this will load key information from the network. We want to use the local initialization instead using loadMintFromCache

packages/core/test/unit/WalletRestoreService.test.ts
```diff
@@ -174,10 +178,11 @@ describe('WalletRestoreService', () => {
         ready: 2,
         spent: 1,
       });
-      expect(proofService.createOutputsAndIncrementCounters).toHaveBeenCalledWith(mintUrl, {
-        keep: 0,
-        send: 74, // 50 + 25 - 1 fee
-      });
+      // saveProofs should have been called with the results from wallet.ops.send
+      expect(proofService.saveProofs).toHaveBeenCalledTimes(1);
```
Egge21M: These changes reflect the new behavior of then sweep method, which is not correct I fear. So once that is changed back, these might have to be reverted as well.

packages/core/operations/send/SendOperationService.ts
```diff
@@ -552,9 +558,8 @@ export class SendOperationService {
               );
 
               // Swap to reclaim
-              const { keep } = await wallet.send(reclaimAmount, sendProofs, {
-                outputData: { keep: outputResult.keep, send: [] },
-              });
+              const reclaimProofs = mapCoreProofsToProofs(sendProofs);
```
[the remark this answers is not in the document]
TheRealCheebs: removed

packages/core/test/unit/MeltQuoteService.test.ts
```diff
@@ -312,10 +348,7 @@ describe('MeltQuoteService.payMeltQuote', () => {
       'spent',
     );
 
-    // Verify meltProofs was called with all selected proofs
-    expect(meltProofsSpy).toHaveBeenCalledWith(quote, selectedProofs);
-
-    // Verify no swap was performed
-    expect(mockProofService.createOutputsAndIncrementCounters).not.toHaveBeenCalled();
+    // Verify meltProofsBolt11 was called with all selected proofs
+    expect(meltProofsBolt11Spy).toHaveBeenCalledWith(quote, selectedProofs);
```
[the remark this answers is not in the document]
TheRealCheebs: Just skipped when I removed some of the calls to createOutputsAndIncreamentCounters from before. It's added back now.

packages/adapter-tests/src/integration.ts
```diff
@@ -1706,25 +1732,27 @@ export async function runIntegrationTests<TRepositories extends Repositories = R
         expect(keypair.publicKeyHex).toBeDefined();
 
         // Create a sender wallet with cashu-ts
-        const senderWallet = new CashuWallet(new CashuMint(mintUrl));
+        const senderWallet = new Wallet(new Mint(mintUrl));
```
robwoodgate: Not incorrect, but less verbose is to pass just the mint url unless using the mint instance elsewhere.

PR #88, opened 2026-02-06
 feat: add multi-unit support (sat, usd, etc.)

4xvgal:
## Summary

  - This PR adds multi-unit support to the coco, allowing wallets to handle multiple currency units (sat, usd
  etc.) per mint. Each mint can have multiple keysets with different units, and this implementation enables filtering proofs
   and balances by unit.

### Key changes:
 - Added unit field to SendOperation and MeltOperation types
 - Updated WalletService to cache wallets per mintUrl:unit combination
 - Added unit-based proof filtering in ProofService
 - Updated all repository implementations to persist the unit field
 - Maintained full backward compatibility with **'sat'** as the default unit
 
---

## Notes to the reviewers

1. **Backward Compatibility:** All unit parameters default to 'sat', so existing code continues to work without changes. When
  reading old database rows without a unit field, it defaults to 'sat'.
2. **No Database Migration Required**: The unit field is optional in row types and defaults to 'sat' when null/undefined.
  Existing data works seamlessly.
3. **API Changes**:
    - WalletApi.send(mintUrl, amount, unit = 'sat') - added optional unit parameter
    - ProofService.selectProofsToSend(mintUrl, amount, unit = 'sat', includeFees) - unit is now the 3rd parameter
    - New methods: getBalancesByUnit(), getSupportedUnits(mintUrl), getReadyProofsByUnit(mintUrl, unit)
4. **Test Coverage**: Added multi-unit test suite in ProofService.test.ts with 4 new tests covering unit-based balance and
  proof filtering.

---

### ADDED

  - WalletApi.getBalancesByUnit() - Get balances grouped by mint and unit
  - WalletApi.getSupportedUnits(mintUrl) - Get list of supported units for a mint
  - ProofService.getBalancesByUnit() - Get all balances grouped by mint and unit
  - ProofService.getReadyProofsByUnit(mintUrl, unit) - Get ready proofs filtered by unit
  - unit field to SendOperation and MeltOperation types
  - Multi-unit test suite in ProofService.test.ts
  
  
 ### MODIFIED

  - WalletService.getWallet(mintUrl, unit = 'sat') - Added optional unit parameter
  - WalletService.getWalletWithActiveKeysetId(mintUrl, unit = 'sat') - Added optional unit parameter
  - WalletService.clearCache(mintUrl, unit?) - Added optional unit parameter
  - WalletService.refreshWallet(mintUrl, unit = 'sat') - Added optional unit parameter
  - ProofService.getBalance(mintUrl, unit = 'sat') - Added optional unit parameter
  - ProofService.selectProofsToSend(mintUrl, amount, unit = 'sat', includeFees) - Added unit parameter
  - WalletApi.send(mintUrl, amount, unit = 'sat') - Added optional unit parameter
  - SendOperationService.init(mintUrl, amount, unit = 'sat') - Added optional unit parameter
  - SendOperationService.send(mintUrl, amount, unit = 'sat') - Added optional unit parameter
  - MeltOperationService.init(mintUrl, method, methodData, unit = 'sat') - Added optional unit parameter
  - All repository implementations (sqlite3, expo-sqlite, indexeddb) to support unit field

### REMOVED

- None


Egge21M (changes requested):
Thank you so much for taking on this challenge! It would be amazing to have this in for the stable v1 release.

packages/core/services/MintQuoteService.ts
```diff
@@ -70,7 +74,8 @@ export class MintQuoteService {
         this.logger?.warn('Mint quote had undefined amount', { mintUrl, quoteId });
         throw new Error('Quote amount undefined');
       }
-      const { wallet } = await this.walletService.getWalletWithActiveKeysetId(mintUrl);
+      const unit = quote.unit || 'sat';
+      const { wallet } = await this.walletService.getWalletWithActiveKeysetId(mintUrl, unit);
```
Egge21M: `quote.unit` is resolved above, but output generation still calls `createOutputsAndIncrementCounters(mintUrl, ...)` without passing unit. In non-`sat` quote redemption this can derive outputs/counters from the default `sat` keyset and fail against the mint (or produce wrong-unit outputs). Please thread `unit` through this output-generation path.

packages/sqlite3/src/repositories/MeltOperationRepository.ts
```diff
@@ -82,6 +84,7 @@ const operationToParams = (operation: MeltOperation): unknown[] => {
     return [
       operation.id,
```
Egge21M: `operationToParams()` now includes `operation.unit`, but the `create()` INSERT column list still omits `unit` while passing `params`. That creates a bindings-count mismatch at runtime when persisting melt operations. Add `unit` to the INSERT columns/placeholders to match params.

Egge21M (changes requested):

Egge21M (changes requested):
Great progress on multi-unit support, but there are a few correctness gaps we need to tackle before merging. One is technically not a bug, but a "behavior" issue that I think should be fixed

Egge21M (changes requested):
I did a more thorough review and noticed that there are still a few places where coco assumes `sat` as unit and things will break. 

We use CDK for our integration tests and it support custom units. I think it would be easy to spot these issues by running the whole integration suite using a custom unit like "USD" against a USD cdk mint once. 

Please let me know if you need help adjusting the integration test to do that

Egge21M (changes requested):

PR #143, opened 2026-04-19
chore: add topological build script for automatic dependency ordering

### LINKED ISSUE #73: Improvement: Proper topological support for build scripts
Our monorepo lacks proper topological support. We do have a internal dependency graph where most packages depend on "core" being built first. Right now we ensure this with command chaining and workflow steps, which is brittle.

We should replace this with an automated topological awareness at some point.

omsant02:
Closes #73

## Problem

The build scripts were manually ordered and brittle:
- Order hardcoded in both `package.json` and `.github/workflows/build-packages.yml`
- Adding a new package requires manually updating both files
- Easy to get wrong order and break builds

## Solution

Added `scripts/build.ts` — a Bun script that:
1. Reads all `package.json` files in `packages/`
2. Builds a dependency graph from `peerDependencies`
3. Sorts packages topologically (dependencies build first)
4. Executes builds in correct order automatically

## Changes

- `scripts/build.ts` — new topological build script
- `package.json` — simplified build script to `bun scripts/build.ts`
- `.github/workflows/build-packages.yml` — simplified to `bun run build`

## Build order output

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -7,11 +7,13 @@ on:
     paths:
       - 'packages/**'
       - 'scripts/build.ts'
+      - 'package.json'
       - '.github/workflows/build-packages.yml'
   pull_request:
     paths:
       - 'packages/**'
       - 'scripts/build.ts'
+      - 'package.json'
       - '.github/workflows/build-packages.yml'
 
 jobs:
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -28,21 +28,7 @@ jobs:
         run: bun install --frozen-lockfile
 
       - name: Build packages in dependency order
-        run: |
-          # Build core first (base dependency)
-          cd packages/core && bun run build && cd ../..
-
-          # Build adapter-tests (depends on core)
-          cd packages/adapter-tests && bun run build && cd ../..
-
-          # Build storage adapters (depend on core and adapter-tests)
-          cd packages/indexeddb && bun run build && cd ../..
-          cd packages/expo-sqlite && bun run build && cd ../..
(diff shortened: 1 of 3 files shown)


Egge21M:
Thanks for the contribution here. Moving the build order logic into a single script is a good cleanup, and the topological approach is a solid direction.

I noticed a few issues before this is ready to merge:

1. The workflow now relies on the root build script, but .github/workflows/build-packages.yml still does not watch the root package.json. That means a future change to the root build entrypoint or workspace config could bypass this workflow entirely.
2. build is now centralized, but typecheck is still hardcoded in the root package.json. Since the workflow also runs bun run typecheck, adding a new package would still require manually updating the root script for CI to cover it.
3. . scripts/build.ts only derives internal build ordering from peerDependencies. In this repo there are also internal workspace relationships in devDependencies such as @cashu/coco-adapter-tests. If the intended contract is that only peerDependencies drive build order, that should be enforced or documented explicitly; otherwise the dependency graph is incomplete and could produce the wrong order later.
  
Happy to take another look after those are addressed.

omsant02:
hi @Egge21M 

Thanks for the review!

Fixed all three points:

1. added package.json to the CI watched paths

2. typecheck now also goes through the topological script so 
   new packages get picked up automatically there too

3. for devDependencies - I tried including them but hit a 
   circular dependency between core and adapter-tests (core 
   devDepends on adapter-tests for tests, adapter-tests 
   devDepends on core to build). This produces wrong order 
   so stuck with peerDependencies only since they already 
   capture all the real build-time relationships correctly.
   Added a comment explaining this + the convention that new 
   packages should declare internal build deps as peerDependencies.

let me know if anything else needs changing!

Egge21M:
Hi @omsant02, thank you for addressing all the issues. This is a great addition and is good to go.

PR #170, opened 2026-05-09
fix: normalize P2PK witness encoding

### LINKED ISSUE #168: Bug: double JSON.stringify on P2PK Witness
P2PK proof witness is encoded 3 times, corrupting the CBOR token.

1. KeyRingService.ts:97 - JSON.stringify when signing (1st)
2. ProofRepository.ts:60 - JSON.stringify on DB save (2nd)
3. @cashu/cashu-ts - JSON.stringify on token encodeing (3rrd)

Those Double stringify from coco impacted 
After token decoding, witness.signaturs is undefined
P2PK signature verification always fails
Tokens cannot be shared with other wallets (no one can decode properly cuz still Stringify'd after decode)

Files to Fix (5 total)
- packages/core/services/KeyRingService.ts:97
- packages/sqlite3/src/repositories/ProofRepository.ts:60
- packages/sqlite-bun/src/repositories/ProofRepository.ts:60
- packages/expo-sqlite/src/repositories/ProofRepository.ts:60
- packages/indexeddb/src/repositories/ProofRepository.ts:48

Fix
- KeyRingService: return witness as object (remove JSON.stringify)
- Repositories: Check if string before encoding (prevent double-encoding)

Small fix (~2 days), but I want to confirm the bug behavior first.

Egge21M:
## Description

This pull request fixes P2PK proof witness encoding so signed proofs no longer produce nested JSON strings when persisted or exported as Cashu tokens.

This pull request resolves #168.

## Problem

`KeyRingService.signProof()` pre-stringified P2PK witnesses, adapters stringified them again for storage, and `cashu-ts` stringified them again during token encoding. Decoded tokens could then expose `witness` as a JSON string literal, causing `witness.signatures` lookups to fail.

## Summary

- Return structured P2PK witness objects from `KeyRingService.signProof()`.
- Add shared witness serialization helpers that normalize object, string, and legacy double-encoded persisted witnesses.
- Use the shared helpers in SQLite3, sqlite-bun, Expo SQLite, and IndexedDB proof repositories.
- Add regression tests for token encoding and persisted witness normalization.
- Include `.changeset/fix-p2pk-witness-encoding.md` for core and all affected proof adapter packages.
- Note: this branch includes the existing `chore: update lockfile` commit and is currently behind `master`.

## Verification

- `bun run --filter='@cashu/coco-core' test -- test/unit/KeyRingService.test.ts test/unit/ProofWitnessSerialization.test.ts`
- `bun run --filter='@cashu/coco-core' build`
- `bun run --filter='@cashu/coco-sqlite' build`
- `bun run --filter='@cashu/coco-sqlite-bun' build`
- `bun run --filter='@cashu/coco-indexeddb' build`
- `bun run --filter='@cashu/coco-expo-sqlite' build`
- `./node_modules/.bin/prettier --check ...`
- `git diff --check`

### DIFF
--- a/packages/core/repositories/ProofWitnessSerialization.ts
+++ b/packages/core/repositories/ProofWitnessSerialization.ts
@@ -1,0 +1,39 @@
+import type { Proof } from '@cashu/cashu-ts';
+
+type ProofWitness = NonNullable<Proof['witness']>;
+
+function normalizeWitnessValue(value: unknown): Proof['witness'] {
+  if (!value) {
+    return undefined;
+  }
+
+  let normalized = value;
+  let lastString = typeof value === 'string' ? value : undefined;
+  while (typeof normalized === 'string') {
+    lastString = normalized;
+    try {
+      normalized = JSON.parse(normalized);
+    } catch {
+      return normalized;
+    }
+  }
+
+  if (normalized && typeof normalized === 'object') {
+    return normalized as ProofWitness;
+  }
+
+  return lastString;
+}
+
+export function normalizeProofWitness(witness: Proof['witness']): Proof['witness'] {
+  return normalizeWitnessValue(witness);
+}
+
+export function stringifyProofWitness(witness: Proof['witness']): string | null {
+  const normalized = normalizeProofWitness(witness);
+  return normalized ? JSON.stringify(normalized) : null;
+}
+
+export function parsePersistedProofWitness(witnessJson: string | null): Proof['witness'] {
+  return normalizeWitnessValue(witnessJson);
+}
--- a/.changeset/fix-p2pk-witness-encoding.md
+++ b/.changeset/fix-p2pk-witness-encoding.md
@@ -1,0 +1,9 @@
+---
+'@cashu/coco-core': patch
+'@cashu/coco-sqlite': patch
+'@cashu/coco-sqlite-bun': patch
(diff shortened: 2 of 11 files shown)


robwoodgate (approved):
Looks good to me.

packages/core/repositories/ProofWitnessSerialization.ts
```diff
@@ -0,0 +1,39 @@
+import type { Proof } from '@cashu/cashu-ts';
+
+type ProofWitness = NonNullable<Proof['witness']>;
+
+function normalizeWitnessValue(value: unknown): Proof['witness'] {
```
Egge21M, line 5: This might have been added later, in 3.5 it doesn't seem to be available. I think for this iteration we should keep the self rolled version as it will also help repair already double and triple stringified witnesses when read from DB

robwoodgate: Yes, sorry, I realized afterwards the CTS function is internal. Also yours undoes any double-encoded, so it's all good.

robwoodgate: CTS exports `getP2PKWitnessSignatures()`, which uses the internal parser. Yours is preferred for now to undo the double-encodings.

PR #174, opened 2026-05-14
Feat: Multi unit support

### LINKED ISSUE #159: feat: multi-unit support
## Summary

  Add first-class support for custom Cashu units in Coco. Today most
  flows still assume `sat` as the default unit, even though units
  already appear in parts of the data model and persistence layer.

  ## Goal

  Coco should let callers work with non-sat units consistently across
  core APIs, React hooks, and persisted wallet/operation state.

  ## API direction

  Update public amount inputs so APIs do not accept a bare numeric
  amount shape like:

  ```ts
  amount: number
   ``` 
  Instead, amount input should support:
  ```ts
  amount: AmountLike | { amount: AmountLike; unit: string }
  ```

  A bare AmountLike can preserve the existing default-unit behavior,
  while the object form makes the unit explicit.

  Internally, Coco should generally treat amount and unit as a coupled
  interface/value rather than passing amount and unit around as loosely
  related fields. That should reduce confusion and prevent accidentally
  mixing a value from one unit with metadata from another.

  ## Scope

  - Core: update operation APIs, services, validation, balance/selection
    logic, quote handling, and history/event data so unit-aware amounts
    are handled consistently.
  - React: expose the same unit-aware input model through hooks and
    derived balance/operation state.
  - Persistence: make sure stored operations, quotes, history, proofs/
    tokens, and migrations preserve unit information and remain
    compatible with existing sat-based data.

  ## Acceptance criteria

  - Existing sat flows continue to work without requiring callers to
    pass { unit: 'sat' } everywhere.
  - Custom-unit mint/send/receive/melt flows do not silently fall back
    to sats.
  - Persisted data round-trips amount and unit together across supported
    adapters.
  - Tests cover both default sat behavior and at least one non-sat
    custom unit.

igbopharaoh:
## Description

This pull request adds first-class support for custom Cashu units across Coco while preserving existing bare-amount `sat` behavior. It makes amount and unit travel together through public APIs, operations, proof storage, balance queries, quote handling, payment requests, React hooks, persistence adapters, migrations, docs, and tests.

This pull request resolves [#159](https://github.com/cashubtc/coco/issues/159).

## Problem

Coco previously assumed `sat` in many internal and public flows. That made custom-unit Cashu usage unsafe because mint/send/receive/melt flows could silently fall back to sats, balances could be interpreted without unit scope, persisted proofs and operations lacked complete unit metadata, and adapter migrations did not consistently round-trip custom units.

## Summary

- Added core unit primitives and parsing helpers in `packages/core/amounts.ts`, including default `sat` behavior, lowercase normalization, object-form amount input, and unit assertions.
- Updated public APIs for mint, send, melt, wallet decode/restore/sweep, balances, and payment requests to accept and preserve units.
- Made `WalletService` unit-scoped, with mint+unit wallet caching and active keyset resolution by unit.
- Added NUT-04/NUT-05 method-unit capability validation while keeping legacy sat quote behavior for mints without capability metadata.
- Made proof storage, proof selection, output generation, inflight checks, reservations, and balance APIs unit-aware.
- Propagated unit through mint, send, receive, melt, restore, sweep, recovery, history, and event payloads.
- Updated `TokenService` to infer unit from proof keysets when token metadata is absent, reject token/keyset unit mismatches, and fallback to `sat` for legacy/unresolvable unitless tokens.
- Kept public `MeltQuoteService` and made quote creation/payment unit-aware.
- Added payment request unit validation so custom-unit requests cannot be funded with sats.
- Updated memory, SQLite3, SQLite-bun, Expo SQLite, and IndexedDB repositories/schemas/migrations to persist proof and operation units.
- Updated React balance context/hooks and operation hooks for unit-scoped balances and custom-unit mint/send/melt inputs.
- Added docs for multi-unit usage and updated existing minting, sending/receiving, payment request, and BIP39 docs.
- Added broad unit, adapter contract, schema migration, React, and optional live integration coverage for custom-unit flows.
- Added `.changeset/many-units-restore.md` for public API and migration behavior.

## Verification

- `bun run --filter='@cashu/coco-core' typecheck`
- `bun run --filter='@cashu/coco-react' typecheck`
- `bun run --filter='@cashu/coco-react' test -- src/lib/hooks/operationHooks.test.tsx src/lib/hooks/useBalances.test.tsx src/lib/providers/root.test.tsx`
- `bun run --filter='@cashu/coco-adapter-tests' build`
- Adapter typechecks for SQLite3, SQLite-bun, Expo SQLite, and IndexedDB
- SQLite-family contract/schema tests passed directly

## Changeset

- Added `.changeset/many-units-restore.md`

### DIFF
--- a/packages/core/api/MeltOpsApi.ts
+++ b/packages/core/api/MeltOpsApi.ts
@@ -62,15 +62,12 @@ export class MeltOpsApi<TSupported extends MeltMethod = DefaultSupportedMeltMeth
    * before committing to the external payment.
    */
   async prepare(input: PrepareMeltInput<TSupported>): Promise<PreparedMeltOperation> {
-    const initOperation =
-      input.unit === undefined
-        ? await this.meltOperationService.init(input.mintUrl, input.method, input.methodData)
-        : await this.meltOperationService.init(
-            input.mintUrl,
-            input.method,
-            input.methodData,
-            input.unit,
-          );
+    const initOperation = await this.meltOperationService.init(
+      input.mintUrl,
+      input.method,
+      input.methodData,
+      input.unit,
+    );
     return this.meltOperationService.prepare(initOperation.id);
   }
 
--- a/packages/core/api/MeltOpsApi.ts
+++ b/packages/core/api/MeltOpsApi.ts
@@ -17,6 +17,8 @@ export type PrepareMeltInput<TSupported extends MeltMethod = DefaultSupportedMel
     method: M;
     /** Method-specific payload required for the selected melt method. */
     methodData: MeltMethodInputData<M>;
+    /** Unit to melt. Defaults to `sat`. */
+    unit?: string;
   };
 }[TSupported];
 
@@ -60,11 +62,15 @@ export class MeltOpsApi<TSupported extends MeltMethod = DefaultSupportedMeltMeth
    * before committing to the external payment.
    */
   async prepare(input: PrepareMeltInput<TSupported>): Promise<PreparedMeltOperation> {
-    const initOperation = await this.meltOperationService.init(
-      input.mintUrl,
-      input.method,
-      input.methodData,
-    );
+    const initOperation =
+      input.unit === undefined
+        ? await this.meltOperationService.init(input.mintUrl, input.method, input.methodData)
+        : await this.meltOperationService.init(
+            input.mintUrl,
+            input.method,
+            input.methodData,
+            input.unit,
+          );
     return this.meltOperationService.prepare(initOperation.id);
   }
 
--- a/packages/core/events/types.ts
+++ b/packages/core/events/types.ts
@@ -24,7 +24,13 @@ export interface CoreEvents {
   };
   'proofs:deleted': { mintUrl: string; secrets: string[] };
   'proofs:wiped': { mintUrl: string; keysetId: string };
-  'proofs:reserved': { mintUrl: string; operationId: string; secrets: string[]; amount: Amount };
+  'proofs:reserved': {
+    mintUrl: string;
+    unit: string;
+    operationId: string;
+    secrets: string[];
+    amount: Amount;
+  };
   'proofs:released': { mintUrl: string; secrets: string[] };
   'melt-quote:created': { mintUrl: string; quoteId: string; quote: MeltQuoteBolt11Response };
   'melt-quote:state-changed': { mintUrl: string; quoteId: string; state: MeltQuoteState };
--- a/packages/core/events/types.ts
+++ b/packages/core/events/types.ts
@@ -1,5 +1,6 @@
-import type { Amount, MeltQuoteBolt11Response, MeltQuoteState, Token } from '@cashu/cashu-ts';
+import type { MeltQuoteBolt11Response, MeltQuoteState, Token } from '@cashu/cashu-ts';
 import type { MeltOperation } from '@core/operations/melt';
+import type { UnitAmount } from '../amounts.ts';
 import type { Counter } from '../models/Counter';
 import type { HistoryEntry } from '../models/History';
 import type { Keyset } from '../models/Keyset';
@@ -26,10 +27,9 @@ export interface CoreEvents {
   'proofs:wiped': { mintUrl: string; keysetId: string };
   'proofs:reserved': {
     mintUrl: string;
-    unit: string;
     operationId: string;
     secrets: string[];
-    amount: Amount;
+    amount: UnitAmount;
   };
   'proofs:released': { mintUrl: string; secrets: string[] };
   'melt-quote:created': { mintUrl: string; quoteId: string; quote: MeltQuoteBolt11Response };
--- a/packages/core/services/MintService.ts
+++ b/packages/core/services/MintService.ts
@@ -261,7 +261,7 @@ export class MintService {
     };
   }
 
-  async assertMintMethodUnitSupported(
+  async assertMethodUnitSupported(
     mintUrl: string,
     nut: 4 | 5,
     method: string,
--- a/packages/core/services/MintService.ts
+++ b/packages/core/services/MintService.ts
@@ -1,4 +1,10 @@
-import { KeysetSyncError, MintFetchError, UnknownMintError } from '../models/Error';
+import { Amount, type AmountLike } from '@cashu/cashu-ts';
+import {
+  KeysetSyncError,
+  MintFetchError,
+  ProofValidationError,
+  UnknownMintError,
+} from '../models/Error';
 import type { Mint } from '../models/Mint';
 import type { Keyset } from '../models/Keyset';
 import type { MintAdapter } from '../infra/MintAdapter';
@@ -8,8 +14,35 @@ import type { CoreEvents } from '../events/types';
 import type { MintInfo } from '../types';
 import type { Logger } from '../logging/Logger.ts';
 import { normalizeMintUrl } from '../utils';
+import { DEFAULT_UNIT, normalizeUnit } from '../amounts.ts';
 
 const MINT_REFRESH_TTL_S = 60 * 5;
+
+export interface MethodUnitCapability {
+  supported: boolean;
+  disabled: boolean;
+  nut: 4 | 5;
+  method: string;
+  unit: string;
+  minAmount?: Amount | null;
+  maxAmount?: Amount | null;
+  options?: unknown;
+  legacySatAllowed?: boolean;
+  reason?: string;
+}
+
+type NutMethodSetting = {
+  method: string;
+  unit: string;
+  min_amount?: AmountLike | null;
+  max_amount?: AmountLike | null;
+  options?: unknown;
+};
+
+type NutMethodSettings = {
+  methods?: NutMethodSetting[];
+  disabled?: boolean;
+};
 
 export class MintService {
   private readonly mintRepo: MintRepository;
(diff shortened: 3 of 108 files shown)


igbopharaoh:
@codex kindly trigger auto review

packages/core/api/MeltOpsApi.ts
```diff
@@ -60,11 +62,15 @@ export class MeltOpsApi<TSupported extends MeltMethod = DefaultSupportedMeltMeth
    * before committing to the external payment.
    */
   async prepare(input: PrepareMeltInput<TSupported>): Promise<PreparedMeltOperation> {
-    const initOperation = await this.meltOperationService.init(
-      input.mintUrl,
-      input.method,
-      input.methodData,
-    );
+    const initOperation =
+      input.unit === undefined
```
Egge21M: In TS omitted arguments become `undefined`, so the ternary should be redundant and you can do

```ts
await this.meltOperationService.init(
            input.mintUrl,
            input.method,
            input.methodData,
            input.unit,
          );
```

instead

packages/core/events/types.ts
```diff
@@ -24,7 +24,13 @@ export interface CoreEvents {
   };
   'proofs:deleted': { mintUrl: string; secrets: string[] };
   'proofs:wiped': { mintUrl: string; keysetId: string };
-  'proofs:reserved': { mintUrl: string; operationId: string; secrets: string[]; amount: Amount };
+  'proofs:reserved': {
+    mintUrl: string;
+    unit: string;
```
Egge21M: I wonder whether we want to use `UnitAmount` here. Wdyt?

packages/core/services/MintService.ts
```diff
@@ -152,6 +185,111 @@ export class MintService {
...
+    nut: 4 | 5,
+    method: string,
+    unit: string,
+  ): Promise<MethodUnitCapability> {
+    const normalizedMintUrl = normalizeMintUrl(mintUrl);
+    const normalizedUnit = normalizeUnit(unit, { defaultUnit: DEFAULT_UNIT });
+    const mintInfo = await this.getMintInfo(normalizedMintUrl);
+    const settings = this.getNutMethodSettings(mintInfo, nut);
+
+    if (!settings || !Array.isArray(settings.methods)) {
```
Egge21M, line 210: I think 
```json
"nuts": {
  "4": {
    "disabled": true
  }
}
```
is a valid info object and would bypass these checks. In that case coco would treat a mint that disabled minting completely as DEFAULT_UNIT and active

Egge21M (changes requested):
Thank you a lot for taking this on. I know it's very large in scope. I think this PR already putss us in a very good place. I did find a couple things, but decided to cut the review short. In the recent dev call we discussed combining `unit` and `amount` in a single object and to always use that internally instead of single amounts and units. That avoid that someone forgets to pass it. 

So at the boundary we want to default to `"sat"`, but internally we should always explicitly require this new struct.

So instead of a signature like `prepare(amount: AmountLike, options?: {unit?: string})` I think we should enforce something like

```ts
export interface UnitAmount {
  unit: string;
  amount: Amount
}

prepare(amount: UnitAmount)
```

This forces us to ALWAYS be explicit about unit + amount combinations internally.

Externally splitting them into two is completely fine so that consumers can simple call `prepare(21)` and get the default `"sat"` fallback behavior

packages/core/events/types.ts
igbopharaoh: Will be better than how I separated it since we're enforcing using unit + amount everywhere

packages/core/services/MintService.ts
igbopharaoh: That's true, this is a new check I added and thought will be better in helping enforce units while checking Mint metadata.
I think I should remove it and allow both sat and other currency unit to go through without mint checks ?

packages/core/api/MeltOpsApi.ts
igbopharaoh: Will make corrections to address this

packages/core/services/MintService.ts
Egge21M: I think the direction is good, however the check needs to handle the case where 04 and 05 are completely disabled regardless of the unit

packages/core/services/ProofService.ts
```diff
@@ -613,19 +755,21 @@ export class ProofService {
    */
   async selectProofsToSend(
     mintUrl: string,
-    amount: AmountLike,
-    includeFees: boolean = true,
+    intent: UnitAmount,
+    options: boolean | ProofSelectionOptions = true,
```
Egge21M: I think renaming this and making it a union at the same time is confusing. I think you changed this because unit was part of options before you switched to combined UnitAmount. Now that there is a single option again I think we should go back to bool in this case

packages/core/services/ProofService.ts
```diff
@@ -467,8 +570,27 @@ export class ProofService {
     );
   }
 
-  private emptyBalanceSnapshot(): BalanceSnapshot {
-    return { spendable: Amount.zero(), reserved: Amount.zero(), total: Amount.zero() };
+  private getSingleBalanceUnit(scope: BalanceQuery | undefined, caller: string): string {
+    const units = normalizeUnitList(scope?.units);
+    if (!units || units.length === 0) {
```
Egge21M: If the caller does not pass a unit filter defaulting to sat is fine, but an empty array should be treated as empty filter and return no results

packages/core/services/MintService.ts
```diff
@@ -152,6 +185,118 @@ export class MintService {
...
+      nut,
+      method,
+      unit: normalizedUnit,
+      minAmount: this.parseOptionalAmount(matchingMethod.min_amount),
+      maxAmount: this.parseOptionalAmount(matchingMethod.max_amount),
+      options: matchingMethod.options,
+    };
+  }
+
+  async assertMintMethodUnitSupported(
```
Egge21M: Please rename to `assertMethodUnitSupported` to avoid confusion about mint vs. melt

packages/core/services/ProofService.ts
```diff
@@ -42,6 +47,21 @@ function countBlankOutputsForAmount(amount: Amount): number {
   return Math.max((value - 1n).toString(2).length, 1);
 }
 
+type ProofSelectionOptions = {
+  includeFees?: boolean;
+};
+
+function normalizeProofSelectionOptions(options?: boolean | ProofSelectionOptions): {
```
Egge21M: Same as https://github.com/cashubtc/coco/pull/174/changes#r3248490263

packages/core/services/ProofService.ts
```diff
@@ -105,25 +127,35 @@ export class ProofService {
     if (inflightProofs.length === 0) {
       return;
     }
-    const batchedByMint: { [mintUrl: string]: CoreProof[] } = {};
+    const batchedByMintAndUnit = new Map<
```
Egge21M, line 118: non-blocking optimization: Technically proof state checks can be done across units. Cashu-ts is the limiting factor here, as it itself is single-unit. We could use the MintAdapter thats used in several other places and build our own cross-unit check function

robwoodgate:
@igbopharaoh @Egge21M - hopefully this will make life easier: https://github.com/cashubtc/cashu-ts/pull/662

Not merged yet, but can push out a 4.4.0 as soon as/if you want it.

packages/core/services/TokenService.ts
```diff
@@ -46,11 +47,238 @@ export class TokenService {
...
+    }
+    if (expectedUnit !== undefined) {
+      assertSameUnit(resolvedUnit, expectedUnit, 'Token');
+    }
+
+    return resolvedUnit;
+  }
+}
+
+function encodedTokenHasExplicitUnit(token: string): boolean {
```
Egge21M: I think this is a great idea, but cashu-ts exposes a helper `getEncodedTokenMetadata` that does essentially this (simple read of cbor encoded data without keyset mapping). I think we should use that instead of re-implementing decoding logic

packages/core/services/WalletService.ts
```diff
@@ -44,93 +43,153 @@ export class WalletService {
     this.authProviderGetter = authProviderGetter;
   }
 
-  async getWallet(mintUrl: string): Promise<Wallet> {
+  async getWallet(mintUrl: string, unit = DEFAULT_UNIT): Promise<Wallet> {
```
Egge21M: All internal services and APIs should expect explicit unit arguments to avoid hidden behavior and confusion

packages/core/services/WalletService.ts
```diff
@@ -44,93 +43,153 @@ export class WalletService {
...
     });
-    this.inFlight.set(mintUrl, promise);
+    this.inFlight.set(cacheKey, promise);
     return promise;
   }
 
-  async getWalletWithActiveKeysetId(mintUrl: string): Promise<{
+  async getWalletWithActiveKeysetId(
+    mintUrl: string,
+    unit = DEFAULT_UNIT,
```
Egge21M: Avoid hidden behavior by defaulting to "sat" internally. We can do this on public APIs to make upgrades less painful for consumers, but internally let's be explciit

Egge21M:
@robwoodgate i think this is a massive QoL and safety update, especially for multi unit wallets like coco is going to be soon.

I'd like to get this PR through the door asap, but I think we should refactor to adopt CTS new type before we release EOM.

igbopharaoh:
@robwoodgate This is really good and would make the approach to multi-unit easier. We just have to switch the Amount type to AmountWithUnit. Great work !!
I'll sync with Egge to know if we'll have it in this pr or in a future release

Egge21M (changes requested):

packages/core/services/ProofService.ts
robwoodgate: CTS `wallet.checkProofsStates()` and `wallet.groupProofsByState()` are unit agnostic. They will return state from ANY unit proofs at the mint the wallet is tied to.

They only info they require is `Proof.secret`... though if you can also pass keyset (`id`) you will be future compatible. (ie: `Pick<Proof, 'secret' | 'id'>`)

So grab whichever wallet instance you have for that mint and use it no problems.

Egge21M: Good point! Then a MintAdapter method is not even required. @igbopharaoh feel free to either add this, or leave it as TODO for a follow up

igbopharaoh: I added a comment for it as a reminder for a follow up.

Egge21M:
@codex please review

Egge21M (approved):

PR #175, opened 2026-05-16
fix(core): keep pending rollback mint-backed

Kelbie:
## Summary

- Keep rollback for pending send operations mint-backed, including exact-match sends where `needsSwap === false`.
- Do not restore original input proofs locally once a send reaches `pending`, because Coco cannot know whether the bearer token has already been shared or spent.
- Surface `PendingSendRollbackError` with `Cannot roll back safely without mint connection. Token might have been spent.` when pending reclaim cannot complete safely.
- Keep `rolling_back` out of the public reclaim retry path for this PR; broader cancel-later/retry behavior is intentionally left for a future design.

## Problem

A wallet can create an ecash token while offline when selected proofs exactly match the requested amount. The original PR tried to make offline cancellation local too, but maintainer feedback clarified that this is unsafe once the operation is `pending`: at that point the token might already have left the device or been redeemed, so rollback must go through the mint to avoid double-spend risk.

## Summary of changes

- Removed the exact-match pending rollback shortcut that restored input proofs locally.
- Restored `SendOpsApi.reclaim()` to only accept `pending` operations.
- Added `PendingSendRollbackError`, exported through the existing core model exports, to preserve the underlying cause while giving clients a clearer message.
- Added focused tests proving exact-match pending rollback uses the mint-backed reclaim path and does not restore proofs locally.
- Added `.changeset/safe-pending-rollbacks.md` for `@cashu/coco-core`.

## Verification

- `bun test packages/core/test/unit/DefaultSendHandler.test.ts packages/core/test/unit/SendOperationService.test.ts packages/core/test/unit/SendOpsApi.test.ts`
- `git diff --check`
- `bun run --filter='@cashu/coco-core' typecheck` still fails on existing integration-test setup noise: `@cashu/coco-adapter-tests` cannot be resolved and an integration-test `@ts-expect-error` is unused. No typecheck errors remain from this patch.

### DIFF
--- a/.changeset/safe-pending-rollbacks.md
+++ b/.changeset/safe-pending-rollbacks.md
@@ -1,0 +1,6 @@
+---
+'@cashu/coco-core': patch
+---
+
+Keep pending send rollback mint-backed and report an explicit unsafe-offline error when reclaim
+cannot complete safely.
--- a/packages/core/api/SendOpsApi.ts
+++ b/packages/core/api/SendOpsApi.ts
@@ -145,14 +145,13 @@ export class SendOpsApi {
    * Attempts to reclaim a pending send operation.
    *
    * This is intended for sends that are already in flight but still support
-   * rollback according to the underlying send method. A previous interrupted
-   * reclaim may leave the operation in `rolling_back`; treat that as retryable.
+   * rollback according to the underlying send method.
    */
   async reclaim(operationId: string): Promise<void> {
     const operation = await this.requireOperation(operationId);
-    if (operation.state !== 'pending' && operation.state !== 'rolling_back') {
+    if (operation.state !== 'pending') {
       throw new Error(
-        `Cannot reclaim operation in state '${operation.state}'. Expected 'pending' or 'rolling_back'.`,
+        `Cannot reclaim operation in state '${operation.state}'. Expected 'pending'.`,
       );
     }
 
--- a/packages/core/api/SendOpsApi.ts
+++ b/packages/core/api/SendOpsApi.ts
@@ -145,13 +145,14 @@ export class SendOpsApi {
    * Attempts to reclaim a pending send operation.
    *
    * This is intended for sends that are already in flight but still support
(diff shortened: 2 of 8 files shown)


Kelbie:
Full disclosure, this was fully vibed with Codex 5.5 High. I did verify this fixed my issue. I can now properly rollback while offline due to these changes. It also helped me clean up my previously corrupted rollbacks too. I was going to make an issue but since I had some working code I'd make a PR so you can have a better understanding of the issue I'm dealing with.

Kelbie:
i think i might be assuming its always safe to rollback ecash sent offline which just isn't true. probably the correct thing to do is to just have proper error handling on my side...

Egge21M:
> i think i might be assuming its always safe to rollback ecash sent offline which just isn't true. probably the correct thing to do is to just have proper error handling on my side...

I think this might be true. The issue is that as soon as `SendOperation` hits `pending`, Coco no longer knows whether the token has been sent out or not. Rolling back a `pending` operation safely requires a mint call, as we might run into double spending issues otherwise. 

Even though it's not beautiful a "Can not roll back safely without mint connection. Token might have been spent" error on the client is the best way.

Egge21M:
@Kelbie do you want to continue working on this? I do think recovering `pending` offline is our of scope, but maybe you could pivot to one of the following:

  - keep exact-match pending rollback blocked offline;
  - possibly improve the error message;
  - keep or add tests proving exact-match pending sends do not restore proofs locally;
  - maybe leave retry support for rolling_back only where rollback recovery is actually safe.

Kelbie:
I'll probably go about improving the error message for now. 

I think the dream would be to have retry logic here, similar to "receive later" feature I've mentioned in an issue. It would be nice to have a "cancel later" so that it eventually rolls back when its safe without users having to be conscious of being online or offline.

I'll pivot and see what I can improve here. I won't do this retry logic yet but if you really like that idea I'm down to try to make that work.

PR #177, opened 2026-05-18
fix(core): unblind restored signatures via toProof() in recoverProofsFromOutputData

### LINKED ISSUE #157: Recovered proofs copy blinded signature instead of unblinding output data
## Problem

`ProofService.recoverProofsFromOutputData()` reconstructs proofs from the mint restore response by copying the blinded signature point into the proof:

```ts
const proof: Proof = {
  id: signature.id,
  amount: signature.amount,
  secret: new TextDecoder().decode(output.secret),
  C: signature.C_,
};
```

`signature.C_` is the blinded signature returned for the blinded output. A valid `Proof.C` must be the unblinded signature point for the output secret.

The sibling change-proof path in the same service already does this correctly via `OutputData.toProof()`:

```ts
output.toProof(sig, { id: keyset.id, keys: keyset.keypairs as Keys })
```

## Verification

This is not related to the cashu-ts v4 migration. The same recovery code path exists on `origin/master`.

I also verified the installed `@cashu/cashu-ts@4.1.0` behavior directly: for the same output data and blinded signature, `signature.C_ !== output.toProof(signature, keyset).C`, while the recovered secret matches. So copying `C_` into `Proof.C` persists a structurally plausible but cryptographically wrong proof.

## Impact

Any path that uses `recoverProofsFromOutputData()` after mint-side signing but before local proof persistence can save bad proofs. Proof-state checks may still appear to pass because they are secret/Y based, but the proof can fail later when spent at the mint.

Known affected recovery surfaces include:

- mint operation recovery
- receive operation recovery
- send rollback/recovery
- P2PK send resurfacing
- melt swap recovery

Current tests do not catch this; at least one recovery test asserts that recovered proof `C` equals the mock `C_`, which encodes the broken behavior.

## Suggested fix

Use `OutputData.toProof()` inside `recoverProofsFromOutputData()` when matching restored signatures back to outputs.

The method will need access to the relevant keyset for each restored signature/output, likely by `signature.id` or `output.blindedMessage.id`, similar to `unblindAndSaveChangeProofs()`.

Add a focused regression test that fails if recovered proof `C` is copied directly from `signature.C_` instead of produced through `OutputData.toProof()`.

KvngMikey:
## Summary

`ProofService.recoverProofsFromOutputData()` was copying the blinded signature point (`C_`) directly into `Proof.C` instead of unblinding it first. This produced structurally plausible but cryptographically invalid proofs that pass secret/Y-based state checks locally but fail when spent at the mint.

## Changes

**`ProofService.ts`**
- Replaced manual `{ id, amount, secret, C: signature.C_ }` construction with `output.toProof(sig, keyset)`
- Keysets are now fetched via `this.mintService.ensureUpdatedMint(mintUrl)` and mapped by `id`, mirroring `unblindAndSaveChangeProofs()`
- If a keyset is missing for a given signature, it is skipped with a warning rather than producing a broken proof

**`ProofService.test.ts`**
- Added a `describe('recoverProofsFromOutputData')` regression test block
- Mocks `OutputData.prototype.toProof` to return `C: 'UNBLINDED_C'` and the restore mock to return `C_: 'BLINDED_C_'`
- Asserts `recovered[0].C === 'UNBLINDED_C'` and `!== 'BLINDED_C_'`
- Extended `mintService` stub with `ensureUpdatedMint` to support the new code path
- This test fails against the old code and passes with the fix

### DIFF
--- a/.changeset/restored-proofs-unblind.md
+++ b/.changeset/restored-proofs-unblind.md
@@ -1,0 +1,5 @@
+---
+'@cashu/coco-core': patch
+---
+
+Unblind restored signatures during proof recovery and reject restored proofs from mismatched keyset units.
--- a/packages/core/services/ProofService.ts
+++ b/packages/core/services/ProofService.ts
@@ -1030,23 +1030,28 @@ export class ProofService {
     // Build blinded messages for restore request
     const blindedMessages = allOutputs.map((o) => o.blindedMessage);
 
+    // Fetch keysets needed to unblind restored signatures
+    const { keysets } = await this.mintService.ensureUpdatedMint(mintUrl);
+    const keysetMap: { [id: string]: Keyset } = {};
+    keysets.forEach((ks) => {
+      keysetMap[ks.id] = ks;
+    });
+
     // Call mint restore endpoint
     const restoreResult = await wallet.mint.restore({ outputs: blindedMessages });
 
-    // Match signatures back to outputs and construct proofs
+    // Match signatures back to outputs and unblind to construct proofs
     const restoredProofs: Proof[] = [];
     for (let i = 0; i < restoreResult.outputs.length; i++) {
       const output = allOutputs.find((o) => o.blindedMessage.B_ === restoreResult.outputs[i]?.B_);
       const signature = restoreResult.signatures[i];
       if (output && signature) {
-        // Construct proof from output data and signature
-        const proof: Proof = {
-          id: signature.id,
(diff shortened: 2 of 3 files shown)


Egge21M:
@codex please review

Egge21M:
Great find of a critical issue. Thank you so much for patching this. I have added a changeset and unit assertions and released a backport of this patch as v1.0.1.

PR #185, opened 2026-05-23
fix: guard against null financial fields in operation hydration

### LINKED ISSUE #156: Repository hydration masks missing prepared operation amounts
## Problem

The persisted operation repositories currently coerce missing amount fields to `0` when hydrating rows for states that should already have prepared data.

Examples in the pre-v4 number-based adapters:

```ts
fee: row.fee ?? 0,
inputAmount: row.inputAmount ?? 0,
```

and for melt operations:

```ts
amount: row.amount ?? 0,
fee_reserve: row.fee_reserve ?? 0,
swap_fee: row.swap_fee ?? 0,
inputAmount: row.inputAmount ?? 0,
```

This is valid for `init` rows only because those rows have not been prepared yet. For prepared-or-later states, these fields are required by the operation model. Defaulting a missing value to zero can hide corrupted or partially written rows and rehydrate them as plausible operations with incorrect financial data.

## Affected surfaces

- Send operation repositories: `fee`, `inputAmount`
- Receive operation repositories: `fee`
- Melt operation repositories: `amount`, `fee_reserve`, `swap_fee`, `inputAmount`
- Applies across persisted adapters, including SQLite-backed adapters and IndexedDB

## Suggested fix

Keep `init` row hydration permissive for fields that are not expected yet, but fail fast for required fields once the state implies prepared data exists.

For example:

```ts
function requireNumber(value: number | null, field: string, operationId: string): number {
  if (value == null) {
    throw new Error(`Invalid operation row ${operationId}: missing ${field}`);
  }
  return value;
}
```

Then use it only after the state check:

```ts
if (row.state === 'init') {
  return { ...base, state: 'init' };
}

const preparedData = {
  fee: requireNumber(row.fee, 'fee', row.id),
  inputAmount: requireNumber(row.inputAmount, 'inputAmount', row.id),
};
```

## Scope

This is a pre-existing repository invariant issue, not part of the cashu-ts v4 amount migration. Track it separately so the v4 migration does not grow in scope.

KvngMikey:
## Summary
This PR fixes a silent data corruption bug in the operation hydration layer of all four adapter packages and centralises the guard helper in each package's `utils.ts`

## Changes
- Added `requireNumber(value, field, operationId)` to each package's `src/utils.ts`; throws `Error: Invalid operation row <id>: missing required field "<field>"` when a required column is NULL or undefined.
- Replaced all `?? 0` fallbacks in the `preparedData` block of `SendOperationRepository`, `ReceiveOperationRepository` and `MeltOperationRepository` across all four adapters with `requireNumber` calls.
- Added a hydration corruption guard test group to sqlite3 and sqlite-bun contract test files covering NULL financial fields for send, receive, and melt operations

### DIFF
--- a/packages/indexeddb/src/utils.ts
+++ b/packages/indexeddb/src/utils.ts
@@ -29,3 +29,14 @@ export function bytesToHex(bytes: Uint8Array): string {
     .map((b) => b.toString(16).padStart(2, '0'))
     .join('');
 }
+
+export function requireNumber(
+  value: string | number | null | undefined,
+  field: string,
+  operationId: string,
+): string | number {
+  if (value == null) {
+    throw new Error(`Invalid operation row ${operationId}: missing required field "${field}"`);
+  }
+  return value;
+}
--- a/packages/indexeddb/src/utils.ts
+++ b/packages/indexeddb/src/utils.ts
@@ -30,7 +30,7 @@ export function bytesToHex(bytes: Uint8Array): string {
     .join('');
 }
 
-export function requireNumber(
+export function assertFieldPresent(
   value: string | number | null | undefined,
   field: string,
   operationId: string,
--- a/packages/sql-storage/src/utils.ts
+++ b/packages/sql-storage/src/utils.ts
@@ -24,3 +24,14 @@ export function bytesToHex(bytes: Uint8Array): string {
     .map((b) => b.toString(16).padStart(2, '0'))
     .join('');
 }
+
+export function requireNumber(
+  value: string | number | null | undefined,
+  field: string,
+  operationId: string,
+): string | number {
+  if (value == null) {
+    throw new Error(`Invalid operation row ${operationId}: missing required field "${field}"`);
+  }
+  return value;
+}
--- a/packages/sql-storage/src/utils.ts
+++ b/packages/sql-storage/src/utils.ts
@@ -25,7 +25,7 @@ export function bytesToHex(bytes: Uint8Array): string {
     .join('');
 }
 
-export function requireNumber(
+export function assertFieldPresent(
   value: string | number | null | undefined,
   field: string,
   operationId: string,
--- a/packages/sqlite3/src/utils.ts
+++ b/packages/sqlite3/src/utils.ts
@@ -29,14 +29,3 @@ export function bytesToHex(bytes: Uint8Array): string {
     .map((b) => b.toString(16).padStart(2, '0'))
     .join('');
 }
-
-export function requireNumber(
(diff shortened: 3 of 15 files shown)


Egge21M:
@codex kindly review

packages/sqlite3/src/utils.ts
```diff
@@ -29,3 +29,14 @@ export function bytesToHex(bytes: Uint8Array): string {
     .map((b) => b.toString(16).padStart(2, '0'))
     .join('');
 }
+
+export function requireNumber(
```
Egge21M: I think this is no longer used, right?

packages/sql-storage/src/utils.ts
```diff
@@ -24,3 +24,14 @@ export function bytesToHex(bytes: Uint8Array): string {
     .map((b) => b.toString(16).padStart(2, '0'))
     .join('');
 }
+
+export function requireNumber(
```
Egge21M: I think this is no longer used, right?

packages/indexeddb/src/utils.ts
```diff
@@ -29,3 +29,14 @@ export function bytesToHex(bytes: Uint8Array): string {
     .map((b) => b.toString(16).padStart(2, '0'))
     .join('');
 }
+
+export function requireNumber(
```
Egge21M: Please consider renaming this, as it does not check whether the value is a number. Also consider `assert` instead of `require`

Egge21M (changes requested):
Thanks for tackling this issue and sorry for the late review. There are some minor issues left. I left comments, mostly related to cleanup. Please also add a changeset.

packages/sqlite3/src/utils.ts
KvngMikey: right.

packages/sql-storage/src/utils.ts
KvngMikey: still used by all 3 operation repos.

packages/indexeddb/src/utils.ts
KvngMikey: this has also been done.

Egge21M (changes requested):
@KvngMikey behavior LGTM, but the changeset you added describes only the last commit, not the change this PR introduces

KvngMikey:
> @KvngMikey behavior LGTM, but the changeset you added describes only the last commit, not the change this PR introduces

@Egge21M , this has been updated now.

Egge21M (approved):

PR #194, opened 2026-06-02
feat(core): persist optional memo on executed send token

### LINKED ISSUE #186: Persist optional memo on executed send tokens
## Problem

  @cashu/cashu-ts Token supports an optional memo, but Coco’s send operation API does not currently provide a way to attach that memo before the token is persisted.

  As a result, callers can only add memo to the returned token object after send.execute(...) completes. That can make the immediate in-memory result look correct, but
  the persisted send operation/history projection still contains the original token without memo.

  ## Relevant Code

  Token supports memo?: string:
```
  // @cashu/cashu-ts
  export type Token = {
    mint: string;
    proofs: Proof[];
    memo?: string;
    unit?: string;
  };
```
  Send history entries already expose the persisted token:
```
  // packages/core/models/History.ts
  export type SendHistoryEntry = OperationHistoryBase & {
    type: 'send';
    amount: Amount;
    state: SendHistoryState;
    token?: Token;
  };
```
  The send history projection reads token from the persisted send operation:
```
  // packages/core/models/History.ts
  export function projectSendOperation(operation: SendOperation): SendHistoryEntry | null {
    // ...
    const token = 'token' in prepared ? prepared.token : undefined;

    return {
      // ...
      ...(token ? { token } : {}),
    };
  }
```
  The public send execution API currently accepts only the operation/id:
```
  // packages/core/api/SendOpsApi.ts
  async execute(
    operationOrId: SendOperation | string,
  ): Promise<{ operation: PendingSendOperation; token: Token }>

  SendOperationService.execute(...) persists the pending operation before callers can mutate the returned token:

  // packages/core/operations/send/SendOperationService.ts
  const result = await handler.execute(ctx);

  if (result.status === 'PENDING') {
    await this.sendOperationRepository.update(result.pending);
    pending = result.pending;
    token = result.token ?? null;
  }
```
  The default and P2PK handlers both create a token and put it on the pending operation:
```
  const token: Token = {
    mint: mintUrl,
    proofs: sendProofs,
    unit: operation.unit,
  };

  const pending: PendingSendOperation = {
    ...operation,
    state: 'pending',
    updatedAt: Date.now(),
    token,
  };
```
  The SQLite repositories persist that operation token through tokenJson:
```
  // packages/expo-sqlite/src/repositories/SendOperationRepository.ts
  // packages/sqlite3/src/repositories/SendOperationRepository.ts
  function serializeToken(operation: SendOperation): string | null {
    const maybeTokenOperation = operation as SendOperation & { token?: unknown };
    return maybeTokenOperation.token ? JSON.stringify(maybeTokenOperation.token) : null;
  }
```
  The history repository later reads tokenJson back into the send history entry:
```
  // packages/expo-sqlite/src/repositories/HistoryRepository.ts
  case 'send': {
    const token = parseToken(row.tokenJson);

    return {
      ...base,
      type: 'send',
      state: row.state as HistoryEntry['state'],
      ...(token ? { token } : {}),
    } as HistoryEntry;
  }
```
  ## Proposed Fix

  Add optional send-token metadata support to the send execution path, starting with memo.

  Suggested API:
```
  // packages/core/api/SendOpsApi.ts
  export interface ExecuteSendOptions {
    memo?: string;
  }

  async execute(
    operationOrId: SendOperation | string,
    options?: ExecuteSendOptions,
  ): Promise<{ operation: PendingSendOperation; token: Token }> {
    const operation = await this.resolveOperation(operationOrId);
    // existing state check...
    return this.sendOperationService.execute(operation, options);
  }
```
  Thread the options into the service:
```
  // packages/core/operations/send/SendOperationService.ts
  export interface ExecuteSendOptions {
    memo?: string;
  }

  function normalizeMemo(memo: string | undefined): string | undefined {
    const trimmed = memo?.trim();
    return trimmed && trimmed.length > 0 ? trimmed : undefined;
  }

  function applyTokenMemo(token: Token | undefined, memo: string | undefined): Token | undefined {
    const normalized = normalizeMemo(memo);
    return token && normalized ? { ...token, memo: normalized } : token;
  }

  async execute(
    operation: PreparedSendOperation,
    options?: ExecuteSendOptions,
  ): Promise<{ operation: PendingSendOperation; token: Token }> {
    // existing setup...

    const result = await handler.execute(ctx);

    if (result.status === 'PENDING') {
      const tokenWithMemo = applyTokenMemo(result.token ?? result.pending.token, options?.memo);

      const pendingWithMemo: PendingSendOperation = tokenWithMemo
        ? { ...result.pending, token: tokenWithMemo }
        : result.pending;

      await this.sendOperationRepository.update(pendingWithMemo);
      pending = pendingWithMemo;
      token = tokenWithMemo ?? null;
    }

    // existing failed/missing-token checks...

    await this.eventBus.emit('send:pending', {
      mintUrl: pending.mintUrl,
      operationId: pending.id,
      operation: pending,
      token,
    });

    return { operation: pending, token };
  }
```
  No database schema change should be needed because tokenJson already persists the full token object.

  ## Acceptance Criteria

  - manager.ops.send.execute(operationOrId, { memo: 'hello' }) returns token.memo === 'hello'.
  - The persisted send operation stores the memo-bearing token in tokenJson.
  - manager.history.getPaginatedHistory(...) returns the send entry with entry.token.memo === 'hello'.
  - Whitespace-only memo is treated as omitted.
  - Existing callers of execute(operationOrId) continue working unchanged.
  - Existing send operations without memo remain unchanged.
  - Both default sends and non-default send handlers, such as P2PK, preserve the memo on the persisted token.

  ## Tests

  Add or update tests in:

  - packages/core/test/unit/SendOpsApi.test.ts
  - packages/core/test/unit/SendOperationService.test.ts
  - packages/core/test/unit/HistoryService.test.ts or repository/history projection tests

  Example coverage:
```
  const prepared = await manager.ops.send.prepare({ mintUrl, amount: 1 });
  const { token } = await manager.ops.send.execute(prepared.id, { memo: 'hello' });

  expect(token.memo).toBe('hello');

  const history = await manager.history.getPaginatedHistory(0, 10);
  const sendEntry = history.find((entry) => entry.type === 'send');

  expect(sendEntry?.token?.memo).toBe('hello');

  Whitespace-only memo should not persist:

  const prepared = await manager.ops.send.prepare({ mintUrl, amount: 1 });
  const { token } = await manager.ops.send.execute(prepared.id, { memo: '   ' });

  expect(token.memo).toBeUndefined();

  const history = await manager.history.getPaginatedHistory(0, 10);
  const sendEntry = history.find((entry) => entry.type === 'send');

  expect(sendEntry?.token?.memo).toBeUndefined();
```
  Also add a service-level test that the send:pending event receives the memo-bearing token, because that event is emitted after persistence and drives the history
  projection path.

KvngMikey:
## Sumary

Adds an optional `memo` option to the send execution path so callers can attach
a memo to the token before it is persisted.

```ts
const { token } = await manager.ops.send.execute(preparedId, { memo: 'hello' });
``` 
The memo is normalized (trimmed, whitespace-only treated as absent) and applied to the token before `sendOperationRepository.update()` so the memo is stored in tokenJson.

- 3 new tests (memo persisted, whitespace omitted, event carries memo-bearing token) and 1 new API-level test (options forwarded).

### DIFF
--- a/packages/core/operations/send/SendOperationService.ts
+++ b/packages/core/operations/send/SendOperationService.ts
@@ -246,9 +246,9 @@ export class SendOperationService {
       };
       await this.sendOperationRepository.update(executing);
 
-      let pending: PendingSendOperation | null = null;
-      let token: Token | null = null;
-      let failed: RolledBackSendOperation | null = null;
+      let pending: PendingSendOperation | undefined;
+      let token: Token | undefined;
+      let failed: RolledBackSendOperation | undefined;
       try {
         const handler = this.handlerProvider.get(operation.method);
         if (!handler) {
@@ -279,17 +279,14 @@ export class SendOperationService {
         const result = await handler.execute(ctx);
 
         if (result.status === 'PENDING') {
-          const tokenWithMemo = this.applyTokenMemo(
-            result.token ?? result.pending.token,
-            options?.memo,
-          );
-          const pendingWithMemo: PendingSendOperation = tokenWithMemo
-            ? { ...result.pending, token: tokenWithMemo }
-            : result.pending;
+          const resolvedToken = options?.memo
+            ? this.applyTokenMemo(result.token!, options.memo)
+            : result.token!;
+          const pendingWithMemo: PendingSendOperation = { ...result.pending, token: resolvedToken };
           // Save the pending operation to the repository
           await this.sendOperationRepository.update(pendingWithMemo);
           pending = pendingWithMemo;
-          token = tokenWithMemo ?? null;
+          token = resolvedToken;
         } else {
           // Handler returned FAILED - persist the terminal result without re-running recovery
           await this.sendOperationRepository.update(result.failed);
@@ -832,14 +829,14 @@ export class SendOperationService {
     return orphanedProofs.length;
   }
 
-  private normalizeMemo(memo: string | undefined): string | undefined {
-    const trimmed = memo?.trim();
-    return trimmed && trimmed.length > 0 ? trimmed : undefined;
+  private normalizeMemo(memo: string): string | undefined {
+    const trimmed = memo.trim();
+    return trimmed.length > 0 ? trimmed : undefined;
   }
 
-  private applyTokenMemo(token: Token | undefined, memo: string | undefined): Token | undefined {
+  private applyTokenMemo(token: Token, memo: string): Token {
     const normalized = this.normalizeMemo(memo);
-    return token && normalized ? { ...token, memo: normalized } : token;
+    return normalized ? { ...token, memo: normalized } : token;
   }
 
   /**
--- a/packages/core/operations/send/SendOperationService.ts
+++ b/packages/core/operations/send/SendOperationService.ts
@@ -280,8 +280,8 @@ export class SendOperationService {
 
         if (result.status === 'PENDING') {
           const resolvedToken = options?.memo
-            ? this.applyTokenMemo(result.token!, options.memo)
-            : result.token!;
+            ? this.applyTokenMemo(result.token, options.memo)
+            : result.token;
           const pendingWithMemo: PendingSendOperation = { ...result.pending, token: resolvedToken };
           // Save the pending operation to the repository
           await this.sendOperationRepository.update(pendingWithMemo);
--- a/packages/core/operations/send/SendOperationService.ts
+++ b/packages/core/operations/send/SendOperationService.ts
@@ -36,7 +36,11 @@ import { MintScopedLock } from '../MintScopedLock';
 import { OperationIdLock } from '../OperationIdLock';
 import { normalizeUnitAmount, type UnitAmount } from '../../amounts.ts';
 
+/**
+ * Options applied when a prepared send operation is executed.
+ */
 export interface ExecuteSendOptions {
+  /** Optional memo to persist on the shareable token. Whitespace-only memos are ignored. */
   memo?: string;
 }
 
@@ -221,6 +225,8 @@ export class SendOperationService {
   /**
    * Execute the prepared operation.
    * Performs the swap (if needed) and creates the token.
+   * If a memo is provided, trims it and persists it on the token before saving the
+   * pending operation. Whitespace-only memos are omitted.
    *
    * If execution fails after transitioning to 'executing' state,
    * automatically attempts to recover the operation.
--- a/packages/core/operations/send/SendOperationService.ts
+++ b/packages/core/operations/send/SendOperationService.ts
@@ -36,6 +36,10 @@ import { MintScopedLock } from '../MintScopedLock';
 import { OperationIdLock } from '../OperationIdLock';
 import { normalizeUnitAmount, type UnitAmount } from '../../amounts.ts';
 
+export interface ExecuteSendOptions {
+  memo?: string;
+}
+
 /**
  * Service that manages send operations as sagas.
  *
@@ -226,6 +230,7 @@ export class SendOperationService {
    */
   async execute(
     operation: PreparedSendOperation,
+    options?: ExecuteSendOptions,
   ): Promise<{ operation: PendingSendOperation; token: Token }> {
     if (!this.handlerProvider) {
       throw new Error('SendHandlerProvider is required');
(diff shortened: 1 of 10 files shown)


packages/core/operations/send/SendOperationService.ts
```diff
@@ -226,6 +230,7 @@ export class SendOperationService {
    */
   async execute(
     operation: PreparedSendOperation,
+    options?: ExecuteSendOptions,
```
Egge21M, line 239: Whats the rationale behind adding memo to execute, but not to prepare?

packages/core/operations/send/SendOperationService.ts
```diff
@@ -274,10 +279,17 @@ export class SendOperationService {
         const result = await handler.execute(ctx);
 
         if (result.status === 'PENDING') {
+          const tokenWithMemo = this.applyTokenMemo(
```
Egge21M: There is much type juggling going on here, even though a execute call that returns `"PENDING"` will always include a token. However the signature does not reflect that. The reason for this is that we used the same ReturnType for regular execution and recovery. I think instead of dancing around this, you should split the ExecutionResult in two types (one for regular -> token required, one for recovery -> token optional)

Egge21M (changes requested):
Thank you for adding this feature. It mostly LGTM, but I think the readability could be improved if we slightly refactor the return types involved.

packages/core/operations/send/SendOperationService.ts
```diff
@@ -226,6 +230,7 @@ export class SendOperationService {
    */
   async execute(
     operation: PreparedSendOperation,
+    options?: ExecuteSendOptions,
   ): Promise<{ operation: PendingSendOperation; token: Token }> {
```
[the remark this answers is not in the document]
KvngMikey: fixed.

packages/core/operations/send/SendOperationService.ts
```diff
@@ -274,10 +279,17 @@ export class SendOperationService {
...
+          const pendingWithMemo: PendingSendOperation = tokenWithMemo
+            ? { ...result.pending, token: tokenWithMemo }
+            : result.pending;
           // Save the pending operation to the repository
-          await this.sendOperationRepository.update(result.pending);
-          pending = result.pending;
-          token = result.token ?? null;
+          await this.sendOperationRepository.update(pendingWithMemo);
+          pending = pendingWithMemo;
+          token = tokenWithMemo ?? null;
```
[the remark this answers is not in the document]
KvngMikey: fixed

packages/core/operations/send/SendOperationService.ts
```diff
@@ -820,6 +832,16 @@ export class SendOperationService {
...
 
+  private normalizeMemo(memo: string | undefined): string | undefined {
+    const trimmed = memo?.trim();
+    return trimmed && trimmed.length > 0 ? trimmed : undefined;
+  }
+
+  private applyTokenMemo(token: Token | undefined, memo: string | undefined): Token | undefined {
+    const normalized = this.normalizeMemo(memo);
+    return token && normalized ? { ...token, memo: normalized } : token;
+  }
```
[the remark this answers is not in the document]
KvngMikey: fixed

packages/core/operations/send/SendOperationService.ts
```diff
@@ -274,10 +279,17 @@ export class SendOperationService {
         const result = await handler.execute(ctx);
 
         if (result.status === 'PENDING') {
+          const tokenWithMemo = this.applyTokenMemo(
+            result.token ?? result.pending.token,
+            options?.memo,
+          );
+          const pendingWithMemo: PendingSendOperation = tokenWithMemo
+            ? { ...result.pending, token: tokenWithMemo }
+            : result.pending;
```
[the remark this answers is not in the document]
KvngMikey: fixed

packages/core/operations/send/SendOperationService.ts
KvngMikey: memo belongs on the `Token` which only exists after `execute`, `prepare` reserves proofs and builds outputs but doesn't produce a token yet so attaching memo at prepare time would mean carrying it as dead state through the operation until execute.

packages/core/operations/send/SendOperationService.ts
KvngMikey: fixed.
i looked at splitting into `ExecuteResult (token required)` and `ExecutionResult (token optional)` but that introduced a duplicate type for a one field difference, which was repetitive and didn't look right to me. i went with the approach that keeps a single `ExecutionResult` using `result.token!` in the PENDING branch (which every real handler satisfies), and retains the runtime guard.

Egge21M: I think this split is worth doing after looking at the service path again.

  The two call sites have different contracts: regular `execute()` must produce a token for the caller, while `recoverExecuting()` may
  legitimately recover a pending operation without being able to reconstruct the token. Keeping one shared `ExecutionResult` makes `token`
  optional everywhere, which then forces `result.token!` in the normal execute path and hides the invariant we actually rely on.

  I’d model that directly:

  - handler `execute()` returns a pending result with `token: Token`
  - handler `recoverExecuting()` returns a pending result with `token?: Token`

  That removes the non-null assertion in `SendOperationService.execute()` and makes future handlers fail at compile time if they don’t return
  the token from normal execution. The small type duplication seems preferable to carrying the recovery-only optionality into the main execute
  path.

Egge21M:
Thanks for taking another look @KvngMikey. I added a new comment to a resolved-thread, because TBH I think this would be the cleaner solution: https://github.com/cashubtc/coco/pull/194#discussion_r3453647504

Please also add a changeset to this, so the feature is added to our changelog

KvngMikey:
@Egge21M , Awesome, all have been addressed.

Egge21M (approved):

PR #307, opened 2026-07-07
fix(core): serialize receive prepare per mint to prevent NUT-13 counter collisions

2100wan:
## Problem

Receiving ecash intermittently fails with:

```
Failed to persist proofs for 1 keyset group(s) [<keysetId>]
```

The per-mint `MintScopedLock` was introduced for **proof selection**: the send and melt services select wallet-owned input proofs, so they acquire it to avoid selecting the same proofs concurrently. Receive has no proof selection — it swaps the proofs carried in the incoming token — so it was excluded.

However, the lock is held across the whole `prepare()`, which also performs the shared, **non-atomic** NUT-13 counter `read → derive → increment` in `createOutputsAndIncrementCounters`. Receive consumes that same per-keyset counter. With no lock, two receive operations on the same mint — e.g. the same gift-wrapped token delivered more than once, or two tokens arriving at the same time — read the same counter value, derive identical blinded outputs, and the second write collides on the unique proof key, producing the error above.

`MintOperationService`, which also has no proof selection, was later given the same lock; receive is the one counter-consuming operation still missing it.

## Fix

- `ReceiveOperationService` now takes the shared `MintScopedLock` and acquires it in `prepare()` around the counter-consuming section, mirroring `SendOperationService`.
- `Manager` wires the same shared lock instance it already passes to the send/melt/mint services.

The lock is an optional last constructor parameter (defaults to a fresh instance), so the change is backward compatible.

## Test

Adds a regression test that runs two concurrent `prepare()` calls for two different tokens on the same mint and asserts the counter-consuming section is never entered concurrently. It fails without the lock (the two prepares overlap) and passes with it.

## Verification

- `bun test packages/core/test/unit` — 1000 passing
- `bun run typecheck` — no new errors introduced by this change

> Note: like the existing lock, this is an in-memory (single-runtime) guard and does not coordinate across processes.

### DIFF
--- a/.changeset/serialize-receive-prepare.md
+++ b/.changeset/serialize-receive-prepare.md
@@ -1,0 +1,5 @@
+---
+'@cashu/coco-core': patch
+---
+
+Serialize concurrent receive `prepare()` per mint with the shared `MintScopedLock`. Receive was left out of the mint-level lock added for proof selection, but it shares the same non-atomic NUT-13 counter derivation, so two concurrent receives on one mint could read the same counter and derive colliding deterministic outputs, failing with "Failed to persist proofs".
--- a/packages/core/Manager.ts
+++ b/packages/core/Manager.ts
@@ -890,6 +890,9 @@ export class Manager {
       walletRestoreLogger,
     );
 
+    // One shared instance across every counter-consuming operation service so that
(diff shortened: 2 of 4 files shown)


Egge21M:
Great catch! Thanks!

Issue #286, opened 2026-07-03
PRD: Support Payment Request P2PK Requirements


Egge21M:
> *This was generated by AI during triage.*

Related issue: #258

## Problem Statement

Coco can read and pay NUT-18 payment requests, and it already has a P2PK send method for producing
NUT-11 locked ecash. However, when a NUT-18 payment request includes a `nut10` spending condition,
Coco does not currently use that condition while paying the request. A receiver can ask for ecash
locked to their P2PK condition, but a Coco payer will prepare an ordinary send instead of delivering
proofs that satisfy the receiver-declared Payment Request P2PK Requirement.

This is risky because the payer application may believe it has paid the request correctly while the
receiver gets proofs that do not match the requested spending condition. It is also unsafe to rely on
mint behavior when the selected mint does not advertise NUT-11 support, because unsupported mints may
treat P2PK secrets as regular anyone-can-spend proofs.

## Solution

Coco should support payer-side Payment Request P2PK Requirements. When Coco reads a NUT-18 payment
request with `nut10.kind === "P2PK"`, it should expose a normalized P2PK requirement for caller
inspection, filter actionable `payableMints` to mints that can satisfy the request, and prepare the
payment as a P2PK send whose locked outputs encode the requested NUT-11 condition.

The implementation should enrich Coco's general P2PK send method data rather than adding a
payment-request-only shortcut. Payment requests should use the `cashu-ts`
`PaymentRequest.toP2PKOptions()` helper from `@cashu/cashu-ts` 4.6.1 or newer to translate the
NUT-18 `nut10` object into the same structured P2PK send target that ordinary P2PK sends can use.
Malformed or unsupported requirements should fail clearly before funds move. Although the helper can
also represent HTLC-style locking as P2PK options, HTLC payment request support remains out of scope
for this PRD.

Delivery behavior should remain unchanged. Inband payment requests return the resulting token for the
application to deliver, HTTP payment requests POST the resulting token, and Nostr delivery remains
owned by a transport plugin.

## User Stories

1. As a Coco payer, I want a payment request's P2PK requirement to be honored, so that I deliver
   ecash the receiver can spend.
2. As a Coco payer, I want payment request parsing to show that a payment requires P2PK locking, so
   that my application can display the security condition before payment.
3. As a Coco payer, I want `payableMints` to exclude mints that cannot satisfy a P2PK payment
   request, so that mint selection only offers actionable choices.
[...]

PR #309, opened 2026-07-10
feat(core): Support P2PK payment request sends

igbopharaoh:
## Description

This pull request adds payer-side NUT-18 P2PK payment request support using cashu-ts NUT-10/NUT-11 normalization. #286 

## Problem

Coco could parse and pay normal payment requests, but P2PK requests needed explicit NUT-11 mint support checks, locked-output send preparation, and clear failure behavior for malformed or unsupported spending conditions before funds move.

## Summary

- Adds normalized P2PK payment request requirements, NUT-11 payable-mint filtering, and prepare-time NUT-11 assertions.
- Routes payment request P2PK sends through the general P2PK send handler with structured cashu-ts options.
- Keeps HTTP/inband/Nostr delivery semantics unchanged and documents the new payer-side behavior.
- Adds adapter contract coverage for structured send method persistence.

## Verification

- `bun run --filter='@cashu/coco-core' test -- test/unit/PaymentRequestService.test.ts`
- `bun run --filter='@cashu/coco-core' test -- test/unit/P2pkSendHandler.test.ts test/unit/SendOpsApi.test.ts test/unit/MintService.test.ts`
- `bun run typecheck`
- `bun run docs:build`
- `git diff --check`

## Changeset

- Added `.changeset/p2pk-payment-requests.md`.

### DIFF
--- a/packages/core/operations/send/SendMethodHandler.ts
+++ b/packages/core/operations/send/SendMethodHandler.ts
@@ -16,11 +16,22 @@ import type {
 } from './SendOperation';
 
 /**
- * Registry of supported send methods and their payload shapes.
- * Extend via declaration merging if you need to add methods externally.
+ * Structured P2PK send options accepted by Coco.
  *
- * Future methods may include:
- * - htlc: { hash: string; timeout: number } - HTLC locked tokens
+ * `hashlock` is intentionally unavailable because cashu-ts treats hashlocked
+ * P2PK options as HTLC/NUT-14 data, which this send method does not support.
+ */
+export type P2pkSendOptions = Omit<P2PKOptions, 'hashlock'> & {
+  /** HTLC/NUT-14 hashlocks are out of scope for P2PK sends. */
+  hashlock?: never;
+};
+
+/**
+ * Payload accepted by the P2PK send method.
+ *
+ * `pubkey` is the legacy shorthand for locking outputs to a single public key.
+ * Prefer `options` for full NUT-11 P2PK conditions such as `sigflag`,
+ * multisig tags, locktime, and refund keys.
  */
 export type P2pkSendMethodData =
   | {
@@ -29,11 +40,18 @@ export type P2pkSendMethodData =
       options?: never;
     }
   | {
-      /** Full NUT-11 P2PK options accepted by cashu-ts output builders. */
-      options: P2PKOptions;
+      /** Full NUT-11 P2PK options accepted by Coco output builders. */
+      options: P2pkSendOptions;
       pubkey?: never;
     };
 
+/**
+ * Registry of supported send methods and their payload shapes.
+ * Extend via declaration merging if you need to add methods externally.
+ *
+ * Future methods may include:
+ * - htlc: { hash: string; timeout: number } - HTLC locked tokens
+ */
 export interface SendMethodDefinitions {
   default: Record<string, never>;
   p2pk: P2pkSendMethodData;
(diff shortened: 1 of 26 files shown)


packages/core/operations/send/SendMethodHandler.ts
```diff
@@ -22,9 +22,21 @@ import type {
  * Future methods may include:
  * - htlc: { hash: string; timeout: number } - HTLC locked tokens
  */
+export type P2pkSendMethodData =
```
Egge21M, line 36: This new root-exported type has no JSDoc of its own; the preceding block still describes the registry and declaration-merging interface below. Please document `P2pkSendMethodData` and keep the registry JSDoc attached to `SendMethodDefinitions`

packages/core/services/PaymentRequestService.ts
```diff
@@ -27,13 +31,42 @@ type PaymentRequestTransport =
   | HttpPaymentRequestTransport
   | NostrPaymentRequestTransport;
 
+export type PaymentRequestP2pkRequirement = {
```
Egge21M, line 39: These exported spending-condition types are part of `PaymentRequestsApi.parse()`'s public result but have no JSDoc describing their semantics.

packages/core/services/MintService.ts
```diff
@@ -221,6 +225,25 @@ export class MintService {
     return mint.mintInfo;
   }
 
+  async supportsNut(mintUrl: string, nut: 11): Promise<boolean> {
```
Egge21M, line 236: Please add JSDoc for the new public `supportsNut()` and `assertNutSupported()` methods. Their refresh/throwing behavior and the currently supported NUT set are not obvious from the signatures.

packages/core/services/PaymentRequestService.ts
```diff
@@ -258,12 +312,143 @@ export class PaymentRequestService {
         balance.spendable.greaterThanOrEqual(amount) &&
         (!mintRequirement || mintRequirement.includes(mintUrl))
       ) {
+        if (
+          spendingCondition?.kind === 'P2PK' &&
+          !(await this.mintService.supportsNut(mintUrl, 11))
```
Egge21M: This can throw if a single mint is offline and take the whole call with it. I think we should catch single mint failures and simply exclude them from the request.

Egge21M (changes requested):
Thank you fr taking this on. The direction is great and this is almost good to go. There are a could docs missing and an issue with how cashu-ts' P2PKOptions behave currently. Please take a look at the comments

igbopharaoh:
> Thank you fr taking this on. The direction is great and this is almost good to go. There are a could docs missing and an issue with how cashu-ts' P2PKOptions behave currently. Please take a look at the comments

Thanks for the review, I'll attend to all the comments now

packages/core/operations/send/SendMethodHandler.ts
```diff
@@ -15,6 +15,25 @@ import type {
...
+ */
+export type P2pkSendMethodData =
+  | {
+      /** Legacy/direct shorthand for sending to one P2PK lock key. */
+      pubkey: string;
+      options?: never;
+    }
+  | {
+      /** Full NUT-11 P2PK options accepted by cashu-ts output builders. */
+      options: P2PKOptions;
```
Egge21M: Cashu-TS lets you add a hashLock on a P2PK construction, which turns the resulting data into HTLC (NUT-14) data. HTLC is out-of-scope for now, so we should make sure that consumers can't add it here.

Egge21M (changes requested):
Thanks for the quick changes. Unfortunately I did find one more issue, which is a bit of a gotcha introduced by CTS.

packages/core/operations/send/SendMethodHandler.ts
igbopharaoh: Great, addressing it now

Egge21M (approved):

Issue #399, opened 2026-07-29
feat: adopt cocod as a workspace package and migrate it to Coco v2


Kelbie:
## Context

Sub-plan for #317. cocod ([Egge21M/cocod](https://github.com/Egge21M/cocod)) is a
Bun CLI + daemon wallet built on Coco, currently pinned to the retired unscoped v1
packages (`coco-cashu-core@1.1.2-rc.50`). In a separate repo it always trails npm
releases — e.g. it cannot pick up the `expiry: 0` watcher fix (#319) until a new
rc ships. As a workspace package it builds against `workspace:*` source, doubles
as a real end-to-end consumer of the core API, and stays current by construction.

An earlier all-in-one attempt (Egge21M/cocod#6, +1633/−90) bundled the v2
migration with onchain, BOLT12, and BIP-321 support. This umbrella re-slices that
work into independently mergeable PRs with explicit blocking edges.

## Plan

Each checkbox is one PR; it can merge once its blockers have merged.

- [x] **1. Import cocod as a workspace package on Coco v2** — blocked by: none
      Add `packages/cocod` (`private: true`, unpublished for now) from a snapshot
      of cocod master. Switch to `workspace:*` deps on `@cashu/coco-core` and
      `@cashu/coco-sqlite-bun` and do the minimal v2 migration so every existing
      route and CLI command keeps working: quote-before-prepare, `Amount` values,
      per-mint balances, plugin registration, NPC v3 account API, manager
      disposal on shutdown. X-Cashu endpoints move to `paymentRequests.*` and
      fail closed on `creqB` and NUT-10 conditions the API cannot yet enforce.
      Wire typecheck + `bun test` into CI. Behavior-preserving; no new features.
- [ ] **2. Onchain (NUT-30) send and receive** — blocked by: 1
      `receive onchain` (deposit addresses), `send onchain` melts with
      cheapest-fee default and explicit fee-option selection, HTTP 202 while a
      melt is pending, and `receive onchain list` via method-filtered
      `listPending`, introducing the shared reusable-quote list helpers.
- [ ] **3. BOLT12 (NUT-25) send and receive** — blocked by: 2
      `receive bolt12` (fixed or reusable offers), `send bolt12` offer melts with
      the same 202-pending semantics, and `receive bolt12 list`. Conceptually
      independent of onchain but reuses the quote-list helpers PR 2 introduces.
- [ ] **4. BIP-321 receive URIs** — blocked by: 2
[...]

PR #401, opened 2026-07-29
feat(cocod): import cocod as a private workspace package on Coco v2

Kelbie:
Implements checkbox 1 of #399 (sub-plan of #317): import cocod as a private workspace package and migrate it to the Coco v2 API. The structural decisions below are called out for reviewer sign-off.

## What this does

- Adds `packages/cocod` (`private: true`, unpublished) as a snapshot of [Egge21M/cocod](https://github.com/Egge21M/cocod) master (`c9da575`), minus repo-level files the monorepo already owns. A snapshot rather than a subtree keeps this repo's history clean; full history stays in the source repo.
- Declares `@cashu/coco-core` and `@cashu/coco-sqlite-bun` as `workspace:*` peerDependencies — peers because `scripts/build.ts` derives build/typecheck order from peerDependencies only — so cocod always builds against workspace source and doubles as a real end-to-end consumer of the core API. cocod is also added to the changesets `ignore` list so the release pipeline never version-bumps the private package.
- Migrates every existing route and CLI command to v2: `initializeCoco({plugins})` + NPC v3 account API, canonical quote-before-prepare mint/melt flows, `wallet.balances.byMint()`, `Amount.toNumber()` at every HTTP boundary.
- Wires CI: typecheck is picked up automatically by `scripts/build.ts` (the package's `lint` script was renamed `typecheck`), and a new `cocod_tests.yml` runs the package tests.

This is deliberately the minimum viable import: every hunk is either forced by the v2 API/monorepo, or listed below with its justification. Onchain (NUT-30), BOLT12 (NUT-25), BIP-321, and the QoL items land as the follow-up PRs tracked in #399. `src/cli.ts` and `src/cli-shared.ts` are byte-identical to the source repo apart from formatting.

## The three deliberate changes beyond the forced API port

- **X-Cashu spending conditions are enforced by core, not the daemon.** The earlier out-of-repo attempt (Egge21M/cocod#6) rejected `creqB` and all NUT-10 conditions at the daemon boundary — correct against cashu-ts 4.5 (whose creqB decoder dropped `nut10`) and pre-#309 core (which prepared unlocked proofs for locked requests). Both defects are fixed on master, so the guards are gone: a committed creqB fixture with a P2PK lock is parsed by real core in a test and classified as an enforceable P2PK condition, and the daemon maps core's `unsupported`/`malformed` classification to a 400 before prepare (the daemon's contract for invalid input) instead of a late 500.
- **Shutdown disposes the manager** (raced against a 3s timeout, timer cleared, timeout logged). This migration is what makes the manager own background watchers, processors, and per-account plugin websockets from boot, so releasing them on stop pairs with the acquisition this PR introduces. Upstream provides `Manager.dispose()` for exactly this host role.
- **New tests pin the migration itself** (balance snapshot totals, the canonical quote splits, numeric boundaries, the inband settlement path, and the creqB/P2PK fixture against real core with an in-memory database). The source repo shipped 3 route tests; these are the evidence for the behavior-preservation claim.

## Decisions that deserve review

- **`workspace:*` instead of the repo's exact-literal internal-dep style.** The literal convention exists to satisfy `check-release.ts` for published packages; cocod is private and exempt, and `workspace:*` can never silently resolve to an npm version across release bumps — and, with cocod ignored by changesets, the spec never needs bumping when core versions change. Happy to switch styles if preferred.
- **Tests are colocated in `src/`** to keep the import diff readable (precedent: `packages/sqlite-bun/src/test/`). Glad to move them to `test/unit` in a follow-up if that layout is preferred.
- **No package-level LICENSE.** The repository root already carries the MIT license that governs every package here, so a second copy inside `packages/cocod` is redundant. The source project's authorship stays recorded through the import commit's reference to Egge21M/cocod, where the original notice lives.

## Known behavior delta vs standalone cocod

- `/balance` returns `{}` for an empty wallet (v1 listed known mints at 0) — unavoidable under the v2 balance API. Reported totals are `snapshot.total` (spendable + reserved), the v2 successor of v1's ready-proof sum.
- `/history` and `/events` forward Coco payloads unchanged, matching standalone cocod — treat local daemon access as wallet-sensitive.

## Known wrinkles

- `coco-cashu-plugin-npc@3.x` (external) declares an exact peer on a core RC, so `bun install` prints an unmet-peer warning against workspace core. Install succeeds; follow-up is to re-cut the plugin with a range peer.
- The Coco database migration is one-way; the package README documents backing up `~/.cocod/coco.db` before upgrading from standalone cocod.
- Mint URLs from config are compared against core-normalized `payableMints` without normalization — the same latent trailing-slash mismatch standalone cocod has. Left as-is to stay behavior-preserving; candidate for a small follow-up fix.

## Testing

- [x] `bun run build`, `bun run typecheck` (cocod included), `bun run format:check`
- [x] `bun run --filter='cocod' test` — 20 tests
- [x] `bun test packages/core/test/unit` — untouched core still green (1149 tests)
- [x] Live smoke against testnut in a sandboxed state dir: init → balance → bolt11 mint auto-redeemed by the background watcher → token send/receive round-trip → `x-cashu parse` with creqA and a P2PK creqB → history → clean shutdown through the dispose path
- [x] README install path: `bun link` from `packages/cocod` yields a global `cocod` running workspace source (verified via the resolved bin target and the same live smoke)
- [x] `security:release-age` passes in CI (local bun predates `Bun.JSONC`)

Refs #399, #317


Egge21M (changes requested):
Thanks for taking this on. I think we need to add some small adjustments.

- The build script runs on peerDependencies only, so it will fail to detect that cocod depends on core and sqlite
- cocod is marked private, but changeset doesn't ignore it, so our release pipeline will verison bump cocod as well. Maybe we should ignore it in changesets for now

Kelbie:
Both addressed in b3757f7:

- `@cashu/coco-core` and `@cashu/coco-sqlite-bun` moved to `peerDependencies` (kept `workspace:*` as the spec since the package is private) — `bun run typecheck` now orders cocod last, after core and sqlite-bun.
- `cocod` added to the changesets `ignore` list. Verified with a scratch `changeset version` run: the published packages bump, cocod stays at 0.0.16 and its `workspace:*` specs are left untouched.

Kelbie:
I'm going to run some more AI reviews myself to help find if I overlooked anything.

Egge21M (approved):

PR #402, opened 2026-07-30
feat(core): add durable mint swap orchestration foundation

### LINKED ISSUE #366: feat(core): add durable mint-swap saga foundation
## Context

This issue defines the revised scope for PR #402, the second slice of #364.

The original implementation pursued the right safety goals, but spread Mint Swap concepts into the
Mint and Melt modules, shared repository types, transaction infrastructure, and an event outbox. The
revised design keeps the durable Saga approach while giving each module a narrower responsibility.

Issue #365 and merged PR #400 already provide the protocol and security foundations needed here:
locked BOLT11 quotes, canonical `amountPaid`/`amountIssued` accounting, monotonic quote observations,
NUT-20 signing and recovery, and sensitive-data logging boundaries. PR #402 will build on those
contracts rather than recreate them.

## Proposed design

A Mint Swap is represented by one durable parent operation that coordinates two existing child
Sagas:

```text
MintSwapOperation
  |-- sourceOperationId ------> MeltOperation
  `-- destinationOperationId -> MintOperation
```

The relationship is intentionally one-way:

- Mint Swap owns the parent state, cross-leg accounting, persistence, and recovery policy.
- Melt owns source-proof reservation, payment execution, rollback, and payment recovery.
- Mint owns destination outputs, issuance, NUT-20 signing, and proof recovery.
- Mint and Melt do not store a Mint Swap parent reference or know about Mint Swap phases.

The parent retains its own `id`. A separate `kind: 'mint-swap'` field is unnecessary because the
record already lives in a dedicated Mint Swap repository. What is being removed is the child-side
`{ kind, id }` ownership marker, not the parent/child relationship.

### Exact child identity

The parent stores both intended child operation IDs before either child is prepared. Recovery uses
those exact IDs; quote lookup is only used to detect inconsistent or competing children.

For the BOLT11-only flow, Coco normally prevents multiple Mint operations for one fixed quote within
a single runtime. However, the repository permits siblings and the service lock is process-local, so
quote identity alone is not a durable correlation guarantee.

The only additional child capability needed is a generic, idempotent `ensure prepared` operation
that accepts a caller-provided opaque operation ID. Repeating the same ID and immutable input returns
the existing child; conflicting input returns a typed conflict. Existing standalone preparation can
delegate to the same implementation after generating its own ID.

This capability belongs to Mint and Melt and contains no Mint Swap terminology. I think we do not need a new
generic plan/authorize/dispatch/apply/recover framework.

## Durable lifecycle

```text
preparing
  -> prepared
  -> source_pending
  -> destination_funded
  -> destination_pending
  -> completed
```

| State                 | Meaning                                                                                                                   |
| --------------------- | ------------------------------------------------------------------------------------------------------------------------- |
| `preparing`           | Both quotes and intended child IDs are recorded. Child preparation may be resumed, but source payment is not authorized.  |
| `prepared`            | Both exact children are prepared and the source debit bounds are known. Execution still requires an explicit request.     |
| `source_pending`      | Persisted before source Melt execution. The source child must be executed or reconciled.                                  |
| `destination_funded`  | Canonical source evidence proves payment and the final debit is within the prepared bounds. Progress is now forward-only. |
| `destination_pending` | Persisted before destination issuance. The exact Mint child and outputs must be executed or recovered.                    |
| `completed`           | Canonical quote accounting and locally stored proofs both prove exact destination issuance.                               |
| `cancelled`           | A requested cancellation completed with no source value transferred and all reservations released.                        |
| `failed`              | A deterministic, value-neutral failure is proven.                                                                         |
| `needs_attention`     | Durable evidence is contradictory or incomplete and another automatic economic action would be unsafe.                    |

`preparing` begins only after both canonical quotes exist. It is persisted before child preparation
because preparing a Melt child can reserve source proofs. A crash after that reservation must leave a
parent that knows exactly which child to recover. Quotes or NUT-20 keys created before the parent are
value-neutral and do not require a larger staged-quote workflow.

The planned coordinator will accept a stable parent operation ID before the first quote request.
Retrying the same ID and immutable intent will resume the same operation; different intent will
produce a typed conflict. This prevents a lost API response from silently creating another parent
and another source reservation.

### Source debit

```text
destinationAmount <= minimum <= maximum <= reserved
maximum <= sourceDebitCap                         when supplied

reserved   = source Melt inputAmount
returned   = pre-swap keep amount + final Melt changeAmount
totalFee   = source Melt swap_fee + effectiveFee
finalDebit = reserved - returned
finalDebit = destinationAmount + totalFee
minimum <= finalDebit <= maximum
```

A paid source child without the settlement information needed to prove these equations moves to
`needs_attention`; the parent does not infer or guess missing values.

### Destination completion

```text
canonical amountIssued = locally stored proof amount = requested destination amount
```

Mint child finality alone is insufficient because issuance may be known while the corresponding
proofs are still absent locally. Recovery must restore and verify the exact persisted outputs before
the parent becomes `completed`.

### Cancellation and failure

A cancellation request is persisted while the operation is `preparing`, `prepared`, or
`source_pending`.

- If source payment is `PENDING`, the request remains recorded while reconciliation continues.
- Confirmed `UNPAID` plus released proofs permits `cancelled`.
- `PAID` crosses the forward-only boundary; the operation must continue toward destination recovery.

Neither timeout, quote expiry, a network error, nor retry exhaustion proves a value-neutral failure.

### Time and diagnostics

Parent timestamps use Unix milliseconds and cannot regress when the local clock moves backwards.
Remote protocol timestamps remain part of the canonical quote records.

Persisted retry and attention diagnostics use bounded codes and typed, non-sensitive evidence. They
must not copy raw invoices, quote identifiers, NUT-20 keys, signatures, outputs, or proofs.

## Persistence and recovery

Mint Swap persistence is supplied as a feature-owned capability:

```ts
interface MintSwapPersistence {
  operationRepository: MintSwapOperationRepository;
}

interface MintSwapOperationRepository {
  create(operation: MintSwapOperation): Promise<void>;
  getById(id: string): Promise<MintSwapOperation | null>;
  transition(command: {
    operationId: string;
    expectedState: MintSwapOperationState;
    expectedRevision: number;
    next: MintSwapOperation;
  }): Promise<boolean>;
  listActive(): Promise<MintSwapOperation[]>;
  listDue(now: number, limit: number): Promise<MintSwapOperation[]>;
}
```

The capability remains optional and dormant in #402; existing adapters and applications are not
required to enable Mint Swap. The repository enforces all-time uniqueness for the parent ID, both
normalized quote references, and both child IDs.

### One concurrency contract

Mint Swap follows the transition pattern in [#463 (Send)](https://github.com/cashubtc/coco/pull/463)
and [#462 (Receive)](https://github.com/cashubtc/coco/pull/462), as requested in the
[maintainer review](https://github.com/cashubtc/coco/issues/366#issuecomment-5539291144):

- `readonly revision: number` is internal persistence metadata. New operations start at `0`.
- `transition` atomically matches ID, state, and revision; missing/mismatched records return
  `false` without mutation.
- Persistence increments once per successful transition, including same-state metadata changes.
  Callers do not choose or pre-increment revisions.
- Invalid transitions do not mutate storage; authoritative changes never use blind `update()`.
- Public commands accept identities and intent, not expected revisions. Internal orchestration
  supplies the concurrency guard.

### Shared Wallet transactions

#417 must bind the same Mint Swap repository interface to the existing physical Wallet transaction
alongside child operations and proofs. Related writes—including the parent's revision—commit or
roll back together. Sharing a database through separate transactions is not sufficient.

#418 adds the Mint Swap domain gateway through the existing Wallet runner/module-factory pattern.
It adds neither an independent transaction mechanism nor a rewrite of #402's model/repository
interface. Mint and Melt remain parent-agnostic.

Transactions stay short and local; remote requests happen outside them. The Saga handles async
results and crashes between commits. This supersedes the earlier blanket exclusion of local
parent/child transactions while retaining Saga recovery.

### Asynchronous recovery

Retry scheduling is persisted. An initial Melt `PENDING` response is normal asynchronous progress:
the parent stays `source_pending` and source proofs remain unavailable across polling and restarts.
Later `PAID` requires settlement verification and forward recovery; later `UNPAID` requires child
rollback and proof release before any permitted value-neutral exit. Pending duration or retry
exhaustion never proves failure.

Ambiguous remote calls enter child recovery before another mutation is attempted. Destination
recovery continues without a retry limit after source payment.

Events may accelerate the same reconciliation process, but are not durable state and may be lost,
duplicated, delayed, or reordered. No event outbox is required.

### Concurrency boundary

V1 supports one active Mint Swap processor per persistence store. Current child repositories do not
fence stale workers across multiple Managers, so a parent-level lease would not make multi-runtime
execution safe. Multi-runtime processing requires separate, generic child-Saga fencing and is not a
guarantee of this slice.

## Scope of PR #402

Included:

- the V1 parent discriminated union and parser;
- legal transitions, immutable facts, cancellation intent, retry metadata, and accounting rules;
- the feature-owned repository contract and memory implementation;
- atomic uniqueness and the Send/Receive-compatible revision/transition contract; and
- focused model, persistence, and concurrency tests.

Excluded:

- a durable event outbox;
- Mint-Swap-specific changes to Mint or Melt;
- persistent adapter schemas and migrations;
- coordinator, processor, Manager/API, history, React, or runtime activation; and
- BOLT12, on-chain, multi-source, batching, or multi-unit swaps.

PR #402 does not wait for the complete transaction migration, but its revision and transition
contract must align now. Shared Wallet transaction support is required for #417's binding and
#418's gateway. Atomic NUT-20 key allocation from #461 is needed before concurrent coordinator
quote preparation, not for this dormant model and memory repository.

## Follow-up slices

- **#417:** persistent adapters, migrations, indexes, and participation in the existing physical
  Wallet transaction, with shared parent/child/proof commit and rollback tests.
- **Generic prerequisite:** exact-ID idempotent preparation in Mint and Melt.
- **#418:** existing-runner transaction gateway, preparation, coordinator, reconciliation, retries,
  and crash tests—including repeated `PENDING` followed by both `PAID` and `UNPAID`, after restart
  and with cancellation intent or lost events.
- **#419:** explicit runtime/API activation, cancellation API, waiters, and best-effort events;
  commands take identities/intent, and pending payments are not reported as terminal.
- **#420:** history, React, documentation, and release acceptance, including both delayed Melt
  outcomes throughout user-facing status.

### PR #402 — durable parent foundation

- [ ] Every one of the nine states parses and round-trips, while illegal state/evidence
      combinations are rejected without copying sensitive persisted values into errors.
- [ ] Stable parent identity, immutable preparation intent, both normalized quote references, and
      both preassigned child IDs are represented for idempotent coordination.
- [ ] Cancellation intent, retry scheduling, bounded diagnostics, and non-recursive last-safe
      evidence are represented in V1.
- [ ] Timestamp, source-settlement, debit-cap, value-neutral-exit, and exact destination-completion
      invariants are tested.
- [ ] Parent ID, both role-scoped quote references, and both child IDs are all-time unique.
- [ ] New records start at revision `0`; revision is read-only persistence metadata and each
      successful transition increments it exactly once, including same-state retry updates.
- [ ] The repository exposes
      `transition({ operationId, expectedState, expectedRevision, next })`; it exposes no
      `compareAndSet`, blind authoritative `update`, or caller-owned revision precondition.
- [ ] Missing IDs and state/revision mismatches return `false` before reading `next`; invalid
      candidates throw; neither path mutates the record or revision, and concurrent transitions
      have one winner.
- [ ] Active and due queries are deterministic, due time is inclusive, limit `0` is a no-op, and
      hydrated values are defensive parsed copies.
- [ ] Mint and Melt production modules contain no Mint Swap symbols, ownership fields, phases, or
      imports.
- [ ] The feature-owned persistence capability remains optional and dormant. #402 adds no persistent
      adapter schema, coordinator, processor, Manager/API surface, event outbox, or independent
      transaction mechanism.
- [ ] Core and adapter-test build/typecheck, core unit tests, and the core coverage gate pass.

### Downstream contract handoff

These are contracts established by #402 and tracked in their owning follow-up issues; their eventual
implementation does not block #402 itself from merging.

- [ ] #417 implements the same repository contract in SQLite3, SQLite Bun, Expo SQLite, and
      IndexedDB behind an opt-in `mintSwap.operationRepository` bundle, with parser-validated
      JSON-safe records, lossless decimal `Amount` values, all five identity constraints, portable
      deterministic scans, and fresh/current-production migrations.
- [ ] #417 binds the nested repository to the existing Wallet transaction handle and lifetime. A
      parent transition, child update, proof update, and parent revision commit or roll back together as one
      physical transaction; no feature runner or nested transaction opener is introduced.
- [ ] #417's adapter contract represents asynchronous Melt persistence without implementing
      reconciliation policy: parent `source_pending` with child Melt `pending`; eventual `PAID` as
      child `finalized` plus parent `destination_funded` and settlement evidence; and eventual
      `UNPAID` only after child `rolled_back`, proof release, and a value-neutral parent exit.
- [ ] #417 injects failures between parent, child, and proof writes and proves no partial async Melt
      branch state or revision survives. Scoped nested calls drain failures and reject after the
      transaction attempt closes.
- [ ] The generic Mint/Melt prerequisite provides idempotent preparation by exact caller-provided
      operation ID without adding Mint Swap terminology or ownership to either child domain.
- [ ] #418 adds `MintSwapTransactions` and `ScopedMintSwapCommands` through the existing Wallet
      runner. Public commands accept identity and intent, while orchestration supplies revision
      guards internally and keeps remote I/O outside transactions.
- [ ] #418 proves repeated Melt `PENDING` across polling and restart, followed by both `PAID` and
      `UNPAID`, including cancellation intent and lost events. Pending duration, quote expiry, and
      retry exhaustion never establish a value-neutral failure.
- [ ] #419 owns runtime/API activation and best-effort post-commit events; #420 owns history, React,
      rollout documentation, changeset, and release acceptance for both delayed Melt outcomes.

igbopharaoh:
## Summary

This refactor replaces the original broad orchestration prototype with a dormant Mint Swap parent
foundation that composes Coco's existing Mint and Melt sagas.

- define a nine-state `MintSwapOperation` discriminated union with exact state-specific facts
- parse persisted records through a strict `unknown`-to-domain boundary with defensive `Amount` and
  binary digest reconstruction
- validate legal transitions, immutable intent, monotonic timestamps, value-neutral exits,
  cancellation intent, retry scheduling, and exact cross-leg accounting
- persist immutable source/destination quote references and preassigned child operation IDs while
  leaving detailed child state authoritative in the existing Mint, Melt, quote, key, and proof
  repositories
- add a feature-owned `MintSwapPersistence` contract and memory repository with revision-guarded
  transitions, all-time identity uniqueness, defensive reads, and deterministic active/due scans
- expose the parent model and parser from core and the persistence/transition contracts from the
  adapter entry point

The parser reuses Coco's existing `normalizeMintUrl`, `deserializeAmount`, `MintQuoteRef`, and
`MeltQuoteRef` contracts. Mint Swap adds only the cross-leg state and evidence needed to coordinate
the two existing child sagas.

## State and recovery contract

The parent progresses through:

```text
preparing -> prepared -> source_pending -> destination_funded
          -> destination_pending -> completed
```

`preparing`, `source_pending`, `destination_funded`, and `destination_pending` are automatically
due. `prepared` waits for explicit execution. Terminal states and `needs_attention` are quiescent.

Cancellation or deterministic failure is allowed only while value neutrality is proven. Once the
source payment is proven paid, recovery is forward-only and completion requires locally stored,
verified destination proofs.

## Scope

This PR provides the model, parser, transition policy, repository contract, memory implementation,
exports, and focused tests. The feature remains dormant.

- Persistent adapter schemas and transaction-scoped bindings remain in #417.
- Coordinator, reconciliation, retry execution, and crash-window tests remain in #418.
- Runtime/API activation and post-commit event hints remain in #419.
- History, React integration, user documentation, and release acceptance remain in #420.

No Mint Swap ownership fields or commands are added to Mint or Melt, and no event outbox, processor,
executable Manager/API surface, persistent adapter, or separate transaction runner is introduced
here.

No changeset is included because this slice does not activate executable Mint Swap behavior; release
metadata remains with the activation and acceptance work.

Closes #366.


igbopharaoh:
@codex kindly review

igbopharaoh:
Hello @Egge21M Kindly review the second slice  of mintswaps

packages/core/operations/melt/MeltOperationService.ts
```diff
@@ -271,6 +311,397 @@ export class MeltOperationService {
     }
   }
 
+  /** Build deterministic source work outside the parent transaction without reserving proofs. */
+  async planOwnedPreparation(
```
Egge21M: Is this planning step really necessary? Building a fake ProofService and no-oping part of it is very brittle. Any sideeffect of ProofService could effect or break this call side.

Could this be omitted and planning happening inside the transaction?

packages/core/services/ProofService.ts
```diff
@@ -81,6 +82,26 @@ export class ProofService {
     this.outputDataCreator = outputDataCreator ?? OutputData;
   }
 
+  /**
+   * Bind local proof and counter writes to a repository transaction.
+   *
+   * Events are intentionally suppressed: a composing parent publishes only after its complete
+   * transaction, including child and parent state, has committed.
+   */
+  forTransaction(repositories: RepositoryTransactionScope): ProofService {
```
Egge21M: This is a great concept and maybe the best direction for coco to support real transactions across the board. However this clones the full ProofService which is still very much able to do async work and I/O which must be done outside of the transaction. E.g. walletService misses its cache and calls the mint, or seedGetter is implemented to call a secure storage API.

A better seam would be a smaller submodule that only exposes the transaction save parts

E.g.

```ts
interface TransactionalProofOperations {
  selectProofsToSend(...): Promise<CoreProof[]>;
  reserveProofs(...): Promise<void>;
  releaseProofs(...): Promise<void>;
  setProofState(...): Promise<void>;
  saveProofs(...): Promise<void>;
  createOutputsAndIncrementCounters(...): Promise<SerializedOutputData>;
}

class ProofService {
  forTransaction(
    repositories: RepositoryTransactionScope,
  ): TransactionalProofOperations {
    return new TransactionalProofOperationsImpl({
      proofRepository: repositories.proofRepository,
      counterRepository: repositories.counterRepository,
      outputDataCreator: this.outputDataCreator,
      seedAndWalletMaterial
    });
  }
}
```

Here only the transaction save parts are exposed. seedAndWalletMaterial can be seed and keysets which are usually fetched automatically and in this case need to be fetched explicitly and put into the transaction scope.

Adding transactions is a great contribution, but also an epic on its own. If your PR relies on them (which is good), please add the transaction capabilities in a separate PR

packages/core/operations/mintSwap/MintSwapOperation.ts
```diff
@@ -0,0 +1,1162 @@
...
+  at: number;
+}
+
+export interface MintSwapTerminalFailure {
+  code: string;
+  reason: string;
+  at: number;
+}
+
+export interface MintSwapOperation {
```
Egge21M: Could we model MintSwapOperation as a discriminated union, following the pattern used by MintOperation, MeltOperation, SendOperation, and ReceiveOperation?

Please introduce a shared base, reusable fact groups such as prepared/source-authorized/funded facts, and a state-specific interface for each lifecycle state. Facts required by a state should be required by its type, while facts that do not apply should be absent or typed as never.

The current flat interface permits invalid combinations. For example, a prepared operation can carry a reconciliation-valid settlement without source dispatch authorization and still pass validateMintSwapOperation.

Please retain runtime validation for persisted data and invariants TypeScript cannot prove, accounting equations, timestamp ordering, URL normalization, fingerprints, revision progression, and immutability. Ideally, that persistence seam should parse unknown or a serialized record into MintSwapOperation, rather than accepting a value already typed as valid.

At minimum, TypeScript should reject:
- prepared with settlement or authorization
- source_inflight without source authorization
- destination_funded without settlement
- issuing without destination authorization
- completed without exact issued amount and completedAt
- failed without terminal failure evidence
- needs_attention without attention evidence and its last-safe checkpoint

This seems worth addressing before the coordinator and persistent adapters build more consumers around the current record shape.

packages/core/infra/handlers/melt/BaseQuoteMeltHandler.ts
```diff
@@ -466,6 +470,193 @@ export abstract class BaseQuoteMeltHandler<M extends MeltMethod> implements Melt
...
+    }
+
+    if (operation.parentExecutionPhase !== 'melt_authorized') {
+      throw new Error(`Melt child ${operation.id} is not awaiting a melt result`);
+    }
+    const proofsToMelt = operation.needsSwap
+      ? getSecretsFromSerializedOutputData(operation.swapOutputData!).sendSecrets
+      : operation.inputProofSecrets;
+
+    switch (result.response.state) {
```
Egge21M: This seems to be duplicating the state handling of `handleMeltReponse` just a couple hundred lines below. Please extract / consolidate

packages/core/operations/mint/MintOperationService.ts
```diff
@@ -246,6 +285,284 @@ export class MintOperationService {
     return this.prepareInitOperation(initOperation.id);
   }
 
+  /**
+   * Prepare and persist a parent-owned destination child using transaction-scoped local writes.
+   *
+   * @internal
+   */
+  async planOwnedPreparation(
```
Egge21M: This makes the Mint module directly aware of Mint Swap ownership, persistence capability, and orchestration phases. The same coupling is added to the Melt module, so the parent saga is spreading into both child modules instead of sitting behind one deep interface. Could Mint and Melt expose generic internal plan / authorize / dispatch / apply / recover commands, with only the Mint Swap module knowing about the parent and its repositories? That would keep child protocol behavior reusable and stop future composed operations from adding another set of parent-specific methods.

packages/core/repositories/index.ts
```diff
@@ -364,6 +369,46 @@ export interface PaymentRequestReceiveAttemptRepository {
...
+ * Keeping the repositories together as one capability prevents a runtime from
+ * observing a partially configured parent/outbox persistence boundary.
+ */
+export interface MintSwapRepositoryCapability {
+  mintSwapOperationRepository: MintSwapOperationRepository;
+  operationEventOutboxRepository: OperationEventOutboxRepository;
+}
+
+/** Fail a Mint Swap command before it can reserve value or perform remote I/O. */
+export function requireMintSwapRepositoryCapability(
```
Egge21M: Is the optional capability intended only as a rollout mechanism, or as the permanent architecture? At present Mint Swap awareness enters RepositoriesBase, every transaction scope, and the mint/melt child commands through requireMintSwapRepositoryCapability. That makes a dormant feature part of the repository interface used by unrelated operations. I would prefer the Mint Swap module to own this persistence bundle and validate it once at its external interface, while child commands receive only the repositories required for their local mutation.

packages/core/models/OperationEventOutbox.ts
```diff
@@ -0,0 +1,192 @@
...
+  revision: number;
+  state: MintSwapOperationState;
+  sourceMintUrl: string;
+  destinationMintUrl: string;
+  unit: 'sat';
+  destinationAmount: string;
+  reasonCode?: string;
+}
+
+export interface OperationEventOutboxRecord {
```
Egge21M: This is a great concept that makes event delivery much more reliable. However I would consider this addition out-of-scope for this change. We should keep it, but it should be an independent PR

Egge21M (changes requested):
The dedicated durable parent-saga approach is directionally correct: cross-mint transfers cannot be protocol-atomic, so authorization-before-I/O, persisted recovery state, CAS, deterministic outputs, and an outbox are appropriate foundations.

My concern is where the orchestration seam currently sits. Mint Swap responsibilities spread into the Mint and Melt modules, the shared repository interface, and the internals of every memory repository. This creates a broad, shallow interface that future coordinators and composed operations will need to understand.
I left comments suggesting a deeper Mint Swap module that exclusively owns parent persistence and coordination and generic typed child commands for Mint and Melt/ I think resolving these seams now is important because the persistent adapters, recovery processor, and public interface will otherwise cement them and make later simplification considerably harder.

This PR also sneaks in a several groundwork concepts that could and should have been separate PRs.

The safety concepts are strong, but I’m requesting changes to simplify the architecture and improving type safety before building the remaining slices on top of it.

igbopharaoh:
Appreciate the comprehensive review on this pull request @Egge21M , I'll take my time to go through the individual reviews and make changes or comments where necessary to simplify my approach and make things clearer.

igbopharaoh:
Hello @Egge21M,  I've fully digested the comprehensive review you gave on my pull request. After going through the review, I think the best way to attend to it is to do a total refactor of my approach and break out some sections which will become independent reviewable PRs particularly the transaction foundation and the event outbox foundation. Then this current PR refactored to address all concerns.

I think this will be a cleaner way to attend to this review but I'll like to hear your thoughts on this proposed approach.

Egge21M:
> I think this will be a cleaner way to attend to this review but I'll like to hear your thoughts on this proposed approach.


I fully agree! Both are good standalone concepts

igbopharaoh:
> > I think this will be a cleaner way to attend to this review but I'll like to hear your thoughts on this proposed approach.
> 
> I fully agree! Both are good standalone concepts

Great, I'll proceed with this approach

igbopharaoh:
@codex kindly re review

Egge21M:
@igbopharaoh you tagged for a review, but the PR is still draft. Is this ready for review?

igbopharaoh:
> @igbopharaoh you tagged for a review, but the PR is still draft. Is this ready for review?

Forgot to remove check, apologies.
Addressed it now.

Egge21M (approved):

PR #408, opened 2026-08-02
fix: allocate key derivation indexes atomically

### LINKED ISSUE #404: fix(core): allocate NUT-20 mint-quote derivation indexes atomically
## Context

Follow-up to #365 and #400.

Source TODO:
https://github.com/cashubtc/coco/pull/400/changes#diff-0103091746c9015f4fece02e671808fd1f9b50c0981611512e59e7f2289d9949R44

`KeyRingService.generateKeyPairForPurpose()` currently reads the last persisted
derivation index, increments it in memory, derives the key, and then persists
the result.

## Problem

Concurrent calls can read the same last derivation index before either call
persists its key. Both calls may consequently derive and return the same
NUT-20 key.

NUT-20 mint-quote ownership requires a fresh key for every quote. Reusing a key
also weakens quote isolation and makes concurrent locked-quote creation
dependent on process-local timing.

This invariant belongs in the upstream key-allocation service and repository
contract rather than in `MintBolt11Handler`.

## Scope

- Add an atomic way to reserve the next derivation index for a key purpose.
- Use that allocation path when generating NUT-20 mint-quote keys.
- Preserve separate derivation sequences for each `KeypairPurpose`.
- Ensure an allocated index cannot later be assigned to another key.
- Update the relevant repository implementations and contract tests.
- Document whether gaps left by interrupted or failed generation are expected.

## Acceptance criteria

- [ ] Concurrent NUT-20 key-generation calls always receive distinct derivation indexes.
- [ ] Concurrent calls always return distinct public keys.
- [ ] Allocation is atomic at the persistence boundary and does not rely only on a process-local lock.
- [ ] Derivation indexes remain isolated by key purpose.
- [ ] A reserved index is never reassigned after a failure; intentional gaps are permitted.
- [ ] Existing deterministic derivation and key-recovery behavior remains compatible.
- [ ] Tests exercise concurrent allocation and failure behavior for every applicable repository.

## Out of scope

- Refactoring `MintBolt11Handler` to coordinate key allocation itself.
- Changing the NUT-20 signing or quote-redemption flow.

igbopharaoh:
This pull request resolves #404.

## Problem

Concurrent key generation previously selected `MAX(stored derivation index) + 1` before deriving
and persisting the keypair. Independent Coco sessions sharing storage could therefore observe the
same maximum, derive the same deterministic NUT-20 quote key, and expose duplicate keys to mints.
A process-local mutex would not coordinate separate SQLite connections, processes, IndexedDB
connections, or browser tabs.

## Solution

- Add a required root-only `KeyRingAllocationRepository` capability that commits a unique,
  purpose-scoped derivation index before seed access and key derivation.
- Preserve permanent gaps after later failures or deletion and report BIP32 child-index exhaustion
  through `DerivationIndexExhaustedError`.
- Add a durable SQL high-water table with migration backfill, guarded writer transactions, bounded
  `SQLITE_BUSY` retries, and cross-connection/process coverage for SQLite3, SQLite Bun, and Expo
  SQLite.
- Add IndexedDB schema version 33 with purpose/index lookup and an atomic two-store Dexie
  transaction, including migration and independent-instance browser coverage.
- Add a reusable allocation contract for built-in and third-party adapters, and document the
  transaction, gap, backup, and storage-security semantics.

## Compatibility

This intentionally changes the public adapter contract. Third-party `Repositories`
implementations must provide `reserveNextDerivationIndex(purpose)` on their root keyring
repository. Transaction scopes retain the narrower non-allocating `KeyRingRepository` interface.
The included Changeset marks `@cashu/coco-core` as a major update and versions the affected adapter
packages.

## Changeset

- Added `.changeset/atomic-key-allocation.md` for the core contract and affected adapters.

### DIFF
--- a/.changeset/atomic-key-allocation.md
+++ b/.changeset/atomic-key-allocation.md
@@ -8,10 +8,10 @@
 '@cashu/coco-indexeddb': patch
 ---
 
-Atomically reserve purpose-scoped key derivation indexes before key generation so concurrent P2PK
-and NUT-20 mint-quote calls cannot reuse deterministic keys. Add durable SQL and IndexedDB
-high-water migrations, permanent gap semantics, cross-instance adapter contracts, and explicit
+Atomically derive and persist purpose-scoped keypairs with their durable high-water marks so
+concurrent P2PK and NUT-20 mint-quote calls cannot reuse deterministic keys. Add SQL and IndexedDB
+high-water migrations, rollback-safe cross-instance adapter contracts, and explicit
 derivation-index exhaustion errors.
 
-Custom repository adapters must replace `getLastDerivationIndex` with the required root-only
-`KeyRingAllocationRepository.reserveNextDerivationIndex` capability.
+Custom repository adapters must replace `getLastDerivationIndex` with the required
+`KeyRingRepository.deriveAndPersistKeyPair` transaction boundary.
--- a/.changeset/atomic-key-allocation.md
+++ b/.changeset/atomic-key-allocation.md
@@ -1,0 +1,17 @@
+---
+'@cashu/coco-core': major
+'@cashu/coco-adapter-tests': minor
+'@cashu/coco-sql-storage': patch
+'@cashu/coco-sqlite': patch
+'@cashu/coco-sqlite-bun': patch
+'@cashu/coco-expo-sqlite': patch
+'@cashu/coco-indexeddb': patch
+---
+
+Atomically reserve purpose-scoped key derivation indexes before key generation so concurrent P2PK
+and NUT-20 mint-quote calls cannot reuse deterministic keys. Add durable SQL and IndexedDB
+high-water migrations, permanent gap semantics, cross-instance adapter contracts, and explicit
+derivation-index exhaustion errors.
+
+Custom repository adapters must replace `getLastDerivationIndex` with the required root-only
+`KeyRingAllocationRepository.reserveNextDerivationIndex` capability.
--- a/bun.lock
+++ b/bun.lock
@@ -59,6 +59,7 @@
       "devDependencies": {
         "@cashu/coco-adapter-tests": "1.0.0",
         "@cashu/coco-sql-storage": "1.0.0",
+        "@types/node": "^25.0.9",
       },
       "peerDependencies": {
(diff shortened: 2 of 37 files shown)


Egge21M (changes requested):
Thank you for the thorough implementation and test coverage. After reviewing the design more closely, I think we are enforcing a stronger invariant than we actually need, which has made the solution considerably more complex.
The current approach assumes that an index must be permanently consumed as soon as a key is derived, even if generation later fails. However, in the current call flow, the key is not exposed or used until generateMintQuoteKeyPair() returns. If the transaction fails before that point, the key may have been computed, but it was never used by the wallet or sent to the mint. Regenerating it later would therefore not constitute key reuse.

I believe the simpler design is:

1. Load the seed before opening the transaction.
2. Start a writer transaction.
3. Read the purpose’s durable high-water index.
4. Derive the next key.
5. Save the key and advance the high-water index in the same transaction.
6. Commit.
7. Return the key only after the commit succeeds.

If the transaction rolls back, neither the key nor the new high-water index exists, and no caller has observed the key. If it commits, both are persisted atomically before the key can be used.

We should retain the durable high-water value so deleting a previously committed key cannot cause its index to be reused. But it does not need to advance when seed loading, derivation, or persistence fails.

For SQLite, this should use a writer transaction such as BEGIN IMMEDIATE; IndexedDB can use its normal read/write transaction. Seed loading should remain outside the transaction because the seed getter is asynchronous, while derivation itself is quick and local.

Could we simplify the PR around this model? That would likely remove the separate pre-committed allocation capability, permanent failure gaps, and much of the allocation-specific machinery while still preventing concurrent callers from receiving the same key. It would also keep the transaction semantics behind the existing repository interface.

igbopharaoh:
> Could we simplify the PR around this model? That would likely remove the separate pre-committed allocation capability, permanent failure gaps, and much of the allocation-specific machinery while still preventing concurrent callers from receiving the same key. It would also keep the transaction semantics behind the existing repository interface.

Thank you for the review Egge, I'll refactor this work to follow this guidelines and provide a feedback before end of day.

Egge21M (approved):

PR #417, opened 2026-08-09
refactor(storage): add mint swap persistence adapters

### LINKED ISSUE #367: feat(storage): add mint swap persistence adapters
## Context

This issue defines the revised scope for PR #417, the third slice of #364.

The original implementation pursued the right durability goals, but coupled Mint Swap persistence
to an operation-event outbox, child-side ownership fields, Mint/Melt-specific commands, and a
separate transaction shape. The revised implementation keeps the durable parent Saga from #366 / PR
#402 while binding it to Coco's existing Wallet persistence and transaction contracts.

PR #417 has been rebased onto the latest `master`, which now contains merged PR #402. The old
prototype was kept only as implementation evidence and was not replayed into the new branch.

## Proposed design

Mint Swap persistence is exposed as one dormant, feature-owned capability:

```ts
interface MintSwapPersistence {
  operationRepository: MintSwapOperationRepository;
}
```

The optional `mintSwap?: MintSwapPersistence` bundle is available at the repository root and inside
the existing Wallet transaction scope. SQLite3, SQLite Bun, Expo SQLite, shared SQL storage, and
IndexedDB enable it only when configured with `mintSwap: true`; ordinary consumers remain
source-compatible and the capability is absent by default.

### Repository and stored model

The persistent adapters implement the #402 contract unchanged:

```ts
interface MintSwapOperationRepository {
  create(operation: MintSwapOperation): Promise<void>;
  getById(id: string): Promise<MintSwapOperation | null>;
  transition(command: {
    operationId: string;
    expectedState: MintSwapOperationState;
    expectedRevision: number;
    next: MintSwapOperation;
  }): Promise<boolean>;
  listActive(): Promise<MintSwapOperation[]>;
  listDue(now: number, limit: number): Promise<MintSwapOperation[]>;
}
```

Each adapter stores the complete V1 parent record plus indexed projections for state, revision,
retry scheduling, timestamps, both role-scoped quote identities, and both child operation IDs.
Every `Amount` is serialized as a lossless decimal string, and every hydrated record passes through
`parseMintSwapOperation`.

The parent ID, source quote, destination quote, source child, and destination child remain unique
for all time. Driver uniqueness failures map to one non-sensitive
`MintSwapIdentityConflictError.kind`; raw driver messages and collided values are not exposed.

### Conditional transitions and scans

`transition` checks the stored ID, state, and revision before interpreting the candidate. A mismatch
returns `false` without mutation. Persistence assigns exactly `expectedRevision + 1`, validates the
legal transition, and performs one conditional write. Same-state retry updates increment the
revision and invalidate stale writers.

Active scans order by `createdAt` and raw JavaScript ID order. Due scans include automatic states due
at or before `now`, order by `nextAttemptAt`, `createdAt`, and raw JavaScript ID order, then apply the
limit. Storage collation does not redefine the #402 ordering contract.

## Shared Wallet transaction

The Mint Swap parent repository uses the same physical adapter transaction as the existing Mint,
Melt, and proof repositories:

```text
one Wallet transaction
  |-- MintSwapOperation transition and revision
  |-- exact Mint or Melt child update
  `-- related proof writes
```

These writes commit or roll back as one local write set. The transaction remains short and local;
remote mint requests and live event publication stay outside it.

Because the feature bundle is nested, `TransactionLifetime` recursively binds
`mintSwap.operationRepository`. Started calls are drained before commit, caught or unobserved
failures poison the attempt, and captured repositories or methods reject after commit or rollback.
This extends the existing runner infrastructure; it does not add a Mint Swap-specific runner or
production gateway.

## Schemas and migrations

- SQL migration `039` adds one Mint Swap parent table with five identity constraints and
  state/revision plus state/due indexes.
- IndexedDB schema version `34` adds the equivalent parent store and indexes.
- Fresh databases and upgrades from the current production schema are covered.
- No operation-event outbox table or Mint/Melt parent-ownership or execution-phase columns are
  introduced.
- Databases created from the unreleased prototype receive no compatibility shim.

## Asynchronous Melt persistence

PR #417 proves only the local persistence shapes required by #418:

- Melt `PENDING` keeps the parent `source_pending` and its source proofs reserved while a same-state
  retry update advances the parent revision.
- Later `PAID` commits child finalization, source settlement evidence, proof state, and the parent
  `destination_funded` transition atomically.
- Later `UNPAID` commits child rollback, proof release, and a valid value-neutral parent exit
  atomically.

Failure injection after each participating write proves that no partial branch state or revision
survives rollback. #418 still owns observation, reconciliation policy, remote effects, and the legal
transition selected from the persisted evidence.

## Scope of PR #417

Included:

- the optional core adapter capability, serializer, typed identity conflict, and memory error
  alignment;
- recursive nested transaction-lifetime binding;
- shared SQL and IndexedDB repositories, schemas, migrations, indexes, and exports;
- opt-in runtime bindings for every maintained adapter; and
- shared repository, concurrency, migration, and physical transaction tests.

Excluded:

- an operation-event outbox or child-side Mint Swap ownership;
- exact-ID child preparation, coordinator, reconciliation, cancellation policy, or retry execution;
- remote I/O, Manager/API activation, events, history, React, or user documentation;
- a generic Saga framework or independent transaction mechanism; and
- release metadata, which remains with activation and acceptance work.

## Follow-up slices

- **#418:** exact-child preparation integration, existing-runner transaction gateway, coordinator,
  reconciliation, retries, and crash recovery.
- **#419:** explicit runtime/API activation, cancellation, waiters, and best-effort post-commit
  events.
- **#420:** history, React integration, user documentation, changeset, and release acceptance.

## Acceptance criteria for #417

- [x] The branch is rebased onto current `master` after #402 merged, without retaining prototype
      architecture or duplicating #402 in the diff.
- [x] Mint Swap persistence is absent by default and opt-in at root and Wallet-transaction scope.
- [x] SQLite3, SQLite Bun, Expo SQLite, and IndexedDB pass one shared #402-derived contract.
- [x] All nine states round-trip with lossless amounts and defensive parser validation.
- [x] All five identities are unique with consistent non-sensitive conflicts across drivers.
- [x] Missing and stale guards are mutation-free; concurrent transitions have one winner and one
      persistence-owned revision increment.
- [x] Active and due scans have identical membership and JavaScript ordering across adapters.
- [x] Parent, child, proof, and revision writes commit or roll back in one physical Wallet
      transaction for `PENDING`, later `PAID`, and later `UNPAID` persistence shapes.
- [x] Nested Mint Swap calls follow the existing drain, failure, retry, and closed-scope lifetime
      contract.
- [x] Fresh and upgrade migrations pass with the required indexes and no prototype artifacts.
- [x] Build, typecheck, core unit tests, SQL migration tests, and cross-browser IndexedDB tests pass
      locally.
- [ ] All PR checks pass on the rebased head; the new workflow run is currently in progress.

igbopharaoh:
## Summary

This refactor replaces the original broad persistence prototype with dormant, opt-in Mint Swap
parent persistence that implements the exact PR #402 contract and participates in Coco's existing
Wallet transaction.

- add one optional `mintSwap.operationRepository` capability at repository root and transaction
  scope for SQLite3, SQLite Bun, Expo SQLite, and IndexedDB
- persist the complete V1 parent record with lossless decimal `Amount` values and parser-validated
  reads
- enforce all-time parent, source/destination quote, and source/destination child identities with
  one non-sensitive typed conflict across adapters
- implement guarded transitions with persistence-owned revisions and portable active/due ordering
- add SQL migration `039` and IndexedDB schema version `34`
- recursively bind nested transaction capabilities and add shared adapter contracts for repository,
  migration, concurrency, and atomic parent/child/proof behavior

## Persistence and transaction contract

The adapters implement only `create`, `getById`, guarded `transition`, `listActive`, and `listDue`.
Mint Swap remains absent unless an adapter is explicitly configured with `mintSwap: true`.

Parent transitions and revision increments use the same physical Wallet transaction as exact Mint
or Melt child updates and related proof writes. Successful callbacks commit the complete local write
set; failures roll it back. Remote mint I/O and live events remain outside the transaction.

The shared contract covers every V1 state, lossless hydration, five identity conflicts, stale and
concurrent writers, raw JavaScript ID ordering, nested scope lifetime, and atomic persistence for
Melt `PENDING`, later `PAID`, and later `UNPAID` outcomes.

## Scope

This PR provides the core adapter boundary and serializer, recursive transaction-lifetime binding,
shared SQL and IndexedDB repositories, schemas, migrations, official adapter wiring, exports, and
focused cross-adapter tests. The feature remains dormant.

No changeset is included because this slice does not activate executable Mint Swap behavior;
release metadata remains with the activation and acceptance work.

Builds on merged PR #402. Closes #367.

### DIFF
--- a/packages/indexeddb/src/repositories/MintSwapOperationRepository.ts
+++ b/packages/indexeddb/src/repositories/MintSwapOperationRepository.ts
@@ -1,0 +1,196 @@
+import { Amount } from '@cashu/cashu-ts';
+import Dexie from 'dexie';
+import {
+  assertMintSwapOperationUpdate,
+  getMintSwapOperationDueAt,
+  isTerminalMintSwapState,
+  validateMintSwapOperation,
+  type MintSwapOperation,
+  type MintSwapOperationState,
+  type MintSwapOperationRepository,
+} from '@cashu/coco-core/adapter';
+
+import { IdbDb, type MintSwapOperationRow } from '../lib/db.ts';
+
+const STORE = 'coco_cashu_mint_swap_operations';
+
+export class IdbMintSwapOperationRepository implements MintSwapOperationRepository {
+  constructor(private readonly db: IdbDb) {}
+
+  async create(operation: MintSwapOperation): Promise<void> {
+    validateMintSwapOperation(operation);
+    if (operation.revision !== 0) {
+      throw new Error('New mint swap operation must start at revision 0');
+    }
+    await this.table().add(toRow(operation));
+  }
+
+  async getById(id: string): Promise<MintSwapOperation | null> {
+    const row = await this.table().get(id);
+    return row ? fromRow(row) : null;
+  }
+
+  async getByState(state: MintSwapOperationState): Promise<MintSwapOperation[]> {
+    const rows = await this.table().where('state').equals(state).toArray();
+    return sortRows(rows).map(fromRow);
+  }
+
+  async getActive(): Promise<MintSwapOperation[]> {
+    const rows = await this.table().toArray();
+    return sortRows(rows)
+      .map(fromRow)
+      .filter((operation) => !isTerminalMintSwapState(operation.state));
+  }
+
+  async getDue(now: number, limit: number): Promise<MintSwapOperation[]> {
+    if (!Number.isSafeInteger(now) || now < 0) throw new Error('Due time must be non-negative');
+    if (!Number.isSafeInteger(limit) || limit < 0)
+      throw new Error('Due limit must be non-negative');
+    if (limit === 0) return [];
+    const rows = await this.table()
+      .where('[dueAt+createdAt+id]')
+      .between([Dexie.minKey, Dexie.minKey, Dexie.minKey], [now, Dexie.maxKey, Dexie.maxKey])
+      .limit(limit)
+      .toArray();
+    return rows.map(fromRow);
+  }
+
+  async getByDestinationMintOperationId(id: string): Promise<MintSwapOperation | null> {
+    return this.getByChild('destinationMintOperationId', id);
+  }
+
+  async getBySourceMeltOperationId(id: string): Promise<MintSwapOperation | null> {
+    return this.getByChild('sourceMeltOperationId', id);
+  }
+
+  async compareAndSet(operation: MintSwapOperation, expectedRevision: number): Promise<boolean> {
+    assertNonNegativeSafeInteger(expectedRevision, 'Expected revision');
+    return this.db.runTransaction('rw', [STORE], async () => {
+      const currentRow = await this.table().get(operation.id);
+      if (!currentRow || currentRow.revision !== expectedRevision) return false;
+      const current = fromRow(currentRow);
+      assertMintSwapOperationUpdate(current, operation);
+      await this.table().put(toRow(operation));
+      return true;
+    });
+  }
+
+  private async getByChild(
+    index: 'destinationMintOperationId' | 'sourceMeltOperationId',
+    id: string,
+  ): Promise<MintSwapOperation | null> {
+    const row = await this.table().where(index).equals(id).first();
+    return row ? fromRow(row) : null;
+  }
+
+  private table() {
+    return this.db.table<MintSwapOperationRow, string>(STORE);
+  }
+}
+
+function sortRows(rows: MintSwapOperationRow[]): MintSwapOperationRow[] {
+  return rows.sort(
+    (left, right) => left.createdAt - right.createdAt || left.id.localeCompare(right.id),
+  );
+}
+
+function toRow(operation: MintSwapOperation): MintSwapOperationRow {
+  return {
+    id: operation.id,
+    state: operation.state,
+    revision: operation.revision,
+    sourceMintUrl: operation.sourceMintUrl,
+    destinationMintUrl: operation.destinationMintUrl,
+    destinationMintOperationId: operation.destinationMintOperationId,
+    sourceMeltOperationId: operation.sourceMeltOperationId,
+    dueAt: getMintSwapOperationDueAt(operation) ?? undefined,
+    createdAt: operation.createdAt,
+    updatedAt: operation.updatedAt,
+    recordJson: JSON.stringify(serializeOperation(operation)),
+  };
+}
+
+function serializeOperation(operation: MintSwapOperation): unknown {
+  return {
+    ...operation,
+    destinationAmount: operation.destinationAmount.toString(),
+    preparedPlan: operation.preparedPlan
+      ? mapAmountsToStrings(operation.preparedPlan, [
+          'sourceMeltAmount',
+          'sourceFeeReserve',
+          'sourcePreparationFee',
+          'sourceMeltInputFee',
+          'minimumSourceDebit',
+          'maximumSourceDebit',
+          'reservedSourceAmount',
+        ])
+      : undefined,
+    settlement: operation.settlement
+      ? mapAmountsToStrings(operation.settlement, [
+          'sourcePaymentFee',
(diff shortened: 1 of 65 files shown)


igbopharaoh:
@codex kindly review

packages/core/operations/mintSwap/MintSwapIdentityConflictError.ts
```diff
@@ -0,0 +1,14 @@
+export type MintSwapIdentityKind =
+  | 'parent'
+  | 'source_quote'
+  | 'destination_quote'
+  | 'source_child'
+  | 'destination_child';
+
+/** A non-sensitive persistence conflict for one all-time Mint Swap identity. */
+export class MintSwapIdentityConflictError extends Error {
```
Egge21M, line 9: Question: What is the reasoning behind this granularity? Does the overall feature require it? It adds some complexity in the adapters and I want to see whether it is worth it.

Egge21M (commented):
LGTM! I left a sinle comment on the complexity around returning different conflict errors. I would like to understand what the greater vision for this error is, or if its just QOL for the consumer

packages/core/operations/mintSwap/MintSwapIdentityConflictError.ts
igbopharaoh: Oh this error is mainly returned for db uniqueness checks when a mintswap wants to be created. So rather than returning the actual error I'm using this to coerce all into one specific domain error. Also it makes it easy for the consumer to know what particular section of the mintswap the uniqueness error is coming from.

packages/sql-storage/src/repositories/MintSwapOperationRepository.ts
```diff
@@ -0,0 +1,259 @@
...
+    const parsed = parseMintSwapOperation(operation);
+    if (parsed.revision !== 0) throw new TypeError('Mint Swap creation requires revision zero');
+    const row = toRow(parsed);
+    try {
+      await this.db.run(
+        `INSERT INTO ${TABLE} (${SELECT_COLUMNS}) VALUES (${Array(15).fill('?').join(', ')})`,
+        rowParams(row),
+      );
+    } catch (error) {
+      const kind = await this.findIdentityConflict(parsed);
```
Egge21M: This can misrepresent any error as ConflictError. So if the insert fails for any other reason, but there is also a conflict present the error handler will throw a ConflictError even though the real cause was something else.

In stead we could remove the catch block and use ON CONFLICT DO NOTHING. Then if the insert returns 0 rows, we can run `findIdentiyConflict` and throw the kind error

packages/indexeddb/src/repositories/MintSwapOperationRepository.ts
```diff
@@ -0,0 +1,195 @@
...
+
+export class IdbMintSwapOperationRepository implements MintSwapOperationRepository {
+  constructor(private readonly db: IdbDb) {}
+
+  async create(operation: MintSwapOperation): Promise<void> {
+    const parsed = parseMintSwapOperation(operation);
+    if (parsed.revision !== 0) throw new TypeError('Mint Swap creation requires revision zero');
+    try {
+      await this.table().add(toRow(parsed));
+    } catch (error) {
```
Egge21M, line 30: Same as for SQL. Check if `error instanceof Dexie.ConstraintError` and only run `findIdentityConflict` in that case

Egge21M (changes requested):
Thanks for addressing my concern about the complexity. I think you are right and the complexity is earned in this case. I found one more issue that can potentially misrepresent errors. Left comments with suggestions

igbopharaoh:
Addressed both review comments using an interactive rebase and pushed changes @Egge21M

Egge21M (approved):

PR #436, opened 2026-08-18
feat(cocod): add BOLT12 send and receive

Kelbie:
Implements checkbox 3 of #399: BOLT12 (NUT-25) send and receive for cocod. No core changes — everything here is served by the existing `quotes.mint` / `quotes.melt` / `ops.melt` APIs.

## What this does

- **`receive bolt12 [--amount]`** creates a canonical BOLT12 mint quote and returns the offer. No mint operation is prepared: an offer stays payable after it is paid, so core's quote watcher and `MintOperationProcessor` claim each payment on their own.
- **`receive bolt12 list`** lists the offers that are still payable, each with what the mint has received against it and what this wallet has already claimed.
- **`send bolt12 <offer> [--amount]`** quotes, prepares, and executes a melt, reporting the amount plus the fee reserve. An unsettled melt returns `202` with the operation id and must not be retried — every payment to an offer is a separate invoice.

## Testing

- [x] `bun run build`, `bun run typecheck`, `bun run format:check`
- [x] `bun run --filter='cocod' test` — 41 tests
- [x] `bun test packages/core/test/unit` — 1249 tests, untouched
- [x] Live against testnut in a sandboxed state dir: reusable offer created, paid, and auto-claimed by the background processor with no follow-up command (3822 sat); fixed-amount offer; `receive bolt12 list`
- [ ] `send bolt12` is covered by route tests only — no live payment yet, as it needs a funded wallet and a payable offer

## Follow-ups

- `send bolt12 --amount` only shapes the melt quote request; core pays whatever the mint quotes back. NUT-25 requires the mint to reject a mismatched request, and cdk additionally verifies the answer in `melt_bolt12_quote`. Happy to open an issue for the equivalent check in `MeltBolt12Handler` if that is wanted — it adds a public error type, so it did not belong here.
- A mint that does not advertise `bolt12` surfaces as a `500`, like every other daemon failure. NUT-25 is optional so this is the likeliest failure; a dedicated status could be worth it, but it is a change to the daemon's error contract rather than part of this feature.

Refs #399

### DIFF
--- a/packages/cocod/README.md
+++ b/packages/cocod/README.md
@@ -10,7 +10,7 @@ If you like simple tools: run commands in your terminal, and let the daemon hand
 - Initialize and secure a Cashu wallet
 - Check balances and transaction history
 - Send and receive Cashu tokens
-- Send and receive Lightning payments (BOLT11)
+- Send and receive Lightning payments (BOLT11 invoices and reusable BOLT12 offers)
 - Handle HTTP 402 payments with `X-Cashu`
 - Manage trusted mints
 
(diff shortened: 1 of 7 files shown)


Egge21M:
Superseded by #439

Issue #444, opened 2026-08-28
feat(core): harden critical Wallet transactions through Send and Receive


Egge21M:
## Problem Statement

Critical Wallet changes are currently spread across orchestration services and repository calls.
Proof selection, proof reservation, deterministic counter advancement, keypair derivation,
operation state changes, and result persistence can commit independently. A crash or concurrent
Coco Session can therefore leave partial state, reuse an allocation, overwrite another operation
transition, or require recovery code to repair state that should have committed atomically.

The existing in-memory locks protect only one Coco Session. Existing adapter transactions also
behave differently when code nests transactions or when independent storage connections write at
the same time. The critical Wallet paths need one recognizable transaction design that works with
IndexedDB and the SQLite adapters without moving remote mint I/O into a database transaction.

## Solution

Introduce one composition-root-owned transaction runner behind narrow domain transaction gateways.
Orchestration modules use those gateways for authoritative local transitions and use read-only query
ports for informational reads. Repositories and the raw runner do not appear in orchestration
interfaces.

Each gateway method performs one complete local domain transition in one strong adapter write
transaction. It reloads authoritative state, verifies proof ownership and operation revision,
performs the complete grouped write, and commits or rolls back as one unit. Orchestration performs
asynchronous preflight before the transaction, remote mint I/O after a durable executing state, and
best-effort live event publication after commit.

The first milestone hardens proof operations, deterministic Output Allocation, and Keypair
Allocation. Send and core Receive are the first complete workflows migrated over that foundation as
separate vertical changes. Send covers owned-proof selection and spending. Receive covers incoming
proof processing, P2PK signing with an existing Wallet key, proof creation, Restore, and Operation
Recovery. Keypair generation receives separate cross-adapter transaction contracts because neither
workflow allocates a new keypair.

## User Stories

1. As a Wallet user, I want a crash during Send preparation to leave either one complete prepared
   Send Operation or no Send Operation, so that my proofs do not become stranded.
2. As a Wallet user, I want a crash during Receive preparation to leave either one complete prepared
   Receive Operation or no Receive Operation, so that deterministic outputs remain recoverable.
3. As a Wallet user, I want concurrent Coco Sessions to allocate different deterministic output
   positions, so that they never derive the same output secrets.
4. As a Wallet user, I want concurrent Send Operations to reserve different proofs, so that the same
   Wallet value cannot be promised to two recipients.
5. As a Wallet user, I want a committed Output Allocation to remain permanently used, so that an
[...]

PR #456, opened 2026-08-30
feat(core): add durable event outbox foundation

### LINKED ISSUE #445: feat(core): add a generic durable event outbox foundation
## Problem Statement

Coco saves Wallet state and sends live events in two separate steps. Coco can stop after it saves
the state but before it sends the event. Required local work can then be lost.

Coco can also send the event before it saves the state. A listener can then receive an event for a
change that does not commit.

The current `EventBus` keeps events only in memory. It does not keep events after a restart. It has
no replay cursor. It cannot commit an event with repository data.

PR #402 showed the need for a durable outbox. That design mixed generic outbox code with Mint Swap
data. It also permitted reads of unclaimed event payloads. The new foundation must separate these
concerns.

## Solution

Add a generic durable event outbox for local work. Store an event record in the same transaction as
the related Wallet state change. Both changes must commit, or both changes must roll back.

After the commit, a worker claims one compatible event record. The worker gives the event record to
a registered local consumer. The consumer changes local repository data.

Commit the consumer change and the `published` status in one transaction. If this transaction
fails, neither change commits. The worker can call the consumer again because the prior consumer
change did not commit.

This design gives exactly one committed local effect. It does not give exactly one call of consumer
code. It does not give exactly one effect in an external system.

The foundation does not depend on a specific transaction runner or transaction participant
interface. The host opens the root transaction and supplies an outbox repository that is already
bound to that transaction. This permits integration with #444 or with another host transaction
mechanism without coupling the outbox to either design.

## User Stories

1. As a Wallet user, I want required local work to continue after Coco restarts.
2. As a Wallet user, I want Wallet state and its event record to commit together.
3. As a maintainer, I want concurrent workers to commit a local consumer effect only once.
4. As a maintainer, I want a claim token to stop an old worker from committing.
5. As a maintainer, I want invalid work to stop after a finite number of attempts.
6. As an operator, I want to inspect pending and blocked work without access to event payloads.
7. As an operator, I want finite storage limits for all outbox data.
8. As an adapter author, I want the same outbox behavior in all storage adapters.
9. As a feature author, I want a small transaction-scoped outbox writer.
10. As an application author, I want the current live `EventBus` behavior to stay unchanged.

## Implementation Decisions

- Store a versioned JSON event record. Include its consumer, stream, revision, event type, payload
  version, event time, content hash, and delivery state.
- Send each event record to one consumer. This outbox is not a publish-and-subscribe system.
- Write one non-empty event batch for one domain revision. Seal the batch with its event count and
  hash.
- Accept an identical retry for the same revision. Reject a different batch for the same revision.
- Create event IDs, timestamps, and random values before a retriable transaction callback. Reuse
  these values during a retry.
- Do not add the outbox to `Repositories`, `RepositoriesBase`, or the public Manager API.
- Give the domain gateway a transaction-scoped outbox writer. Do not give it a root transaction
  method.
- Let the host own root transactions and database-conflict retries. An outbox repository uses only
  the supplied transaction-bound storage handle. It does not open or retry a root transaction.
- Let a worker claim only an event contract that it supports. Identify a contract by consumer ID,
  event type, envelope version, and payload version.
- Give each claim a random lease token and an expiry time.
- Require the current lease token for publish, retry, block, and consumer commit operations.
- Read and validate the claim again in the consumer transaction. Check the event contract and
  content hash before the consumer runs.
- Give the consumer only the repositories that it needs for its local effect. Do not give it an
  HTTP client, mint client, `EventBus`, timer, or publish method.
- Commit the local effect and `markPublished` in one transaction.
- Do not use JavaScript cancellation as a safety control. Cancellation is cooperative. Consumer
  code can continue after a lease expires.
- Serialize competing writes. Check the lease token again before commit. Only the current token can
  commit the effect and `published` status.
- Do not count a confirmed database-conflict retry as a delivery failure. The host can retry only
  when it knows that the transaction did not commit.
- Use a finite retry limit and capped exponential backoff. Store only safe error codes and safe
  messages.
- Move invalid work and exhausted work to `blocked`. Require an explicit action to requeue blocked
  work.
- Do not guarantee delivery order. The local effect must be independent or commutative. Otherwise,
  use a `lastAppliedRevision` check in the same transaction.
- Set finite limits for event rows, revision seals, streams, and payload bytes. Check the limits in
  the producer transaction.
- Roll back the related Wallet state change if the outbox is full. Do not save state without its
  required event record.
- Keep published payloads for a defined retention period. Then compact them in bounded batches.
- During compaction, move the stream checkpoint and delete covered rows in one transaction.
- Never compact pending or blocked rows.
- Keep one checkpoint for each retained stream. The checkpoint prevents reuse of a deleted
  revision.
- Store canonical JSON. Do not store proof secrets, full proofs, tokens, private keys, seeds,
  preimages, authentication data, or raw transport data.
- Do not include payloads, raw errors, or stack traces in logs and inspection results.

## Testing Decisions

- Test JSON validation, canonical hashes, batch seals, identical retries, conflicting retries,
  capacity limits, retry delays, blocking, requeue, compaction, and safe logs.
- Add one shared adapter test suite.
- Test producer commit and rollback with a generic host fixture and persistent transaction scopes.
- Cause a failure after a consumer repository write. Verify that the effect and `published` status
  both roll back.
- Run concurrent workers. Include a lease-expiry test. Verify that only one local effect commits.
- Verify that an old lease token cannot publish, retry, block, or commit an effect.
- Verify that a worker cannot claim an unsupported event contract.
- Verify that compaction rejects young, leased, pending, blocked, corrupt, or incomplete data.
- Verify that checkpoint movement and row deletion commit or roll back together.
- Use two storage connections for SQLite3 and Bun SQLite tests.
- Use two Dexie instances and browser Workers for IndexedDB claim and enqueue tests.
- Run native Expo transaction and claim tests before an outbox feature is enabled on Expo. Keep
  native Expo activation disabled until these tests pass.
- Test an upgrade from the prior schema. Keep all existing data during the upgrade.
- Verify that an old IndexedDB session receives `versionchange` and closes. An old client can reopen
  and write stores that it knows, so the host must fence or reload old sessions before producer
  activation.

## Delivery Sequence

1. Agree on the event, claim, consumer, retry, capacity, retention, and compaction contracts.
2. Add the core types, validation, worker, consumer coordinator, and generic host tests.
3. Add SQL and IndexedDB storage implementations.
4. Run the shared adapter, migration, contention, typecheck, and build tests.
5. Add a feature producer and production worker in a separate adoption issue.

Before production use, identify the first local consumer and its exact repository change. Register
that consumer before its producer starts. This order prevents unsupported pending work.

Do not use this outbox if the proposed consumer only calls `EventBus` or an external service. These
calls cannot commit in the same local database transaction.

## Out of Scope

- Mint Swap event types, payloads, producers, and consumers.
- Changes to Manager, `EventBus`, or current `CoreEvents`.
- Public replay, subscriber cursors, fan-out, or strict event order.
- Remote mint calls, HTTP requests, analytics calls, and other external effects.
- Exactly one call of consumer code.
- Exactly one effect in an external system.
- A root transaction runner, a required participant interface, or a nested root transaction.
- Proof, counter, Send, Receive, and operation service changes from #444.
- Automatic deletion of unresolved work.
- A destructive down migration.

## Acceptance Criteria

- Wallet state and its sealed event batch commit or roll back in one physical transaction.
- A supported local effect and the `published` status commit or roll back together.
- A crash, repeated worker call, concurrent worker, or expired lease cannot cause two committed local
  effects.
- Unsupported event contracts stay pending. Compatible code can process them later.
- Invalid or exhausted work becomes `blocked`. Inspection does not expose its payload.
- Compaction limits published row growth. Stream checkpoints continue to prevent duplicate
  revisions.
- Hard limits keep unresolved storage growth finite.
- All supported adapters pass the shared tests. Disable outbox use in an adapter until its required
  native tests pass.
- The foundation does not change a global repository interface, Manager API, or `EventBus`
  behavior.
- The foundation builds and passes its tests without #444. A later feature can adapt any suitable
  host transaction mechanism to the narrow writer and consumer scopes.

## Further Notes

- PR #402 caused this work. This issue is separate from the Mint Swap feature.
- Delivery is at least once when a compatible worker runs. Permanent failures stay visible as
  pending or blocked work.
- Exactly once applies only to the committed local repository effect. Consumer code can run again
  after a rollback.
- An external system needs its own idempotency key or durable delivery protocol.
- Public listeners must read authoritative state after a restart. This outbox does not replay live
  events.

igbopharaoh:
Closes #445.

## Problem

Coco persists authoritative wallet state and emits live in-memory events as separate operations.
If a process, browser tab, mobile app, or daemon stops after the state commit but before the event
is handled, required local follow-up work can be lost. Emitting first has the opposite failure
mode: consumers can observe a transition that later rolls back.

The existing `EventBus` is intentionally a live notification mechanism. It has no durable record,
claim lease, restart recovery, or shared transaction with wallet persistence, so it cannot provide
durable local processing by itself.

## What this PR adds

This PR adds an opt-in, generic durable event outbox foundation without coupling it to Mint Swap,
Manager, `EventBus`, issue #444, or a specific feature transaction runner.

- Versioned JSON-safe event envelopes, canonical payload hashing, and sealed aggregate-revision
  batches.
- Idempotent enqueue with stream revision checks and immutable event intent.
- Contract-filtered claims, lease-token fencing, bounded retries, blocking, requeue, inspection,
  retention, and checkpoint-preserving compaction.
- A bounded publisher and a transactional consumer coordinator for rollback-capable local effects.
- Persisted finite row, revision, stream, and payload-byte capacity limits.
- Public outbox transaction ports for memory, IndexedDB, SQLite3, Bun SQLite, and Expo SQLite.
- Combined Wallet/outbox transaction methods in persistent adapters so authoritative state and
  event intent can share one physical commit.
- Shared repository and host-atomicity conformance coverage through each adapter's public API.
- SQL migration `039_durable_event_outbox` and IndexedDB schema version 34.

The outbox is not added to Coco's global `Repositories` interface. Each participating adapter
exposes the capability explicitly, and each feature remains responsible for opting in.

### 1. Produce state and event intent atomically

The feature creates stable IDs, timestamps, and payload inputs before entering a callback that may
be retried. Inside the adapter's combined transaction, it updates authoritative Wallet state and
enqueues one sealed event batch:

```ts
const batch = createStableEventBatch(command);
const now = clock.now();

await repositories.withDurableEventOutboxTransaction(async (scope) => {
  await scope.mintRepository.addOrUpdateMint(updatedMint);
  await scope.durableEventOutbox.enqueueRevision(batch, now);
});
```

Both writes commit, or both roll back. A semantic retry of the same stream revision returns the
existing event IDs. A different batch for an already sealed revision is rejected.

### 2. Claim only supported work

`DurableEventOutboxPublisher.runOnce(limit)` opens short transactions through
`repositories.durableEventOutbox`. It advertises the exact consumer contracts registered in that
process and claims only compatible events. Each claim receives a worker ID, unique lease token,
and expiry; every later mutation checks that token so an expired worker cannot acknowledge work
owned by a replacement worker.

### 3. Commit the local effect with publication

A consumer created with `createDurableEventConsumerExecutor()` receives a feature-owned
transaction port. Inside that transaction the coordinator:

1. reloads and validates the current claim and immutable content hash;
2. applies the supported rollback-capable local repository effect; and
3. marks the event `published` using the same lease token.

If any step fails, both the local effect and acknowledgement roll back. The callback may be entered
again, but no earlier local effect committed. This is exactly-once **committed local effect**, not
exactly-once function invocation.

### 4. Recover and bound storage

Transient failures are rescheduled with finite backoff. Deterministic, corrupt, or exhausted work
moves to `blocked` and can be inspected or explicitly requeued without exposing payloads or claim
tokens. Published rows can be compacted after retention; a monotonic stream checkpoint remains so
an old revision cannot be recreated. Persisted capacity policy prevents unresolved work from
growing storage without a bound.

## Adapter availability

| Adapter     | Root outbox transactions                  | Combined Wallet/outbox transaction           |
| ----------- | ----------------------------------------- | -------------------------------------------- |
| Memory      | `MemoryDurableEventOutboxTransactionPort` | Host stages other in-memory state explicitly |
| IndexedDB   | `repositories.durableEventOutbox.run()`   | `withDurableEventOutboxTransaction()`        |
| SQLite3     | `repositories.durableEventOutbox.run()`   | `withDurableEventOutboxTransaction()`        |
| Bun SQLite  | `repositories.durableEventOutbox.run()`   | `withDurableEventOutboxTransaction()`        |
| Expo SQLite | `repositories.durableEventOutbox.run()`   | `withDurableEventOutboxTransaction()`        |

SQLite outbox transactions acquire the writer lock with `BEGIN IMMEDIATE` before their first read.
Confirmed `SQLITE_BUSY`/`SQLITE_LOCKED` conflicts may repeat the callback, so callers must capture
stable inputs beforehand and must not perform external effects inside it.

Expo's generic repository capability is enabled. Its transaction path and contention behavior are
covered through the Expo-compatible async/native shim, including `BEGIN IMMEDIATE`. Each concrete
Expo feature must still pass its producer, consumer, and competing-claim tests on supported native
devices before that feature's producer is activated.

## Guarantees and boundaries

When a feature composes the provided transaction scopes correctly, this foundation guarantees:

- authoritative local state and event intent commit or roll back together;
- a supported local consumer effect and `published` acknowledgement commit or roll back together;
- pending and blocked work survives restart;
- stale lease holders cannot mutate work after replacement;
- duplicate revisions cannot silently change semantic content; and
- compaction preserves replay protection while bounding retained rows.

It does **not** make HTTP calls, remote mint operations, analytics, arbitrary callbacks, or
`EventBus` listeners exactly once. Those effects cannot share the local database transaction and
require target-side idempotency or another durable protocol. This PR also does not activate a
producer, scheduler, Manager behavior, or feature event; those belong in later feature-integration
changes after consumers are deployed.

## Rollout notes

- Deploy/register consumers before enabling producers.
- A participating feature must prove its producer state and event batch share one transaction.
- Its consumer must prove its local effect and publication acknowledgement share one transaction.
- Old IndexedDB sessions must be fenced or reloaded before producer activation because schema
  upgrade alone does not prevent older code from writing stores it already knows.
- Stop scheduling and inspect pending/blocked contract versions before rolling back an activated
  producer.

The local Expo custom-unit run completed schema initialization but later timed out against the
local USD mint container; the PR workflow remains the authoritative custom-unit verification.

## Changeset

`.changeset/durable-event-outbox-foundation.md` records minor releases for the new public adapter
capabilities and core API, plus the shared adapter conformance package.


igbopharaoh:
@codex kindly review

igbopharaoh:
@codex kindly re-review

igbopharaoh:
Hello @Egge21M good day, this PR is open for review.
In the meantime, I'm still looking into why CI is failing with `pollingTransport` tests that aren't related to this PR.

igbopharaoh:
Closing this as we prefer to use the sagas to add durable state persistence and the generic outbox is not required

PR #488, opened 2026-09-10
docs: add shared and browser Coco wallet skills

Egge21M:
Adds composable consumer skills for agents building Cashu wallets with Coco. `coco-wallet` owns shared Wallet identity, public API, trust, amount, and recovery guidance. Its task table routes agents to separate references for ecash send, ecash receive, BOLT11, BOLT12, onchain, payment requests, P2PK, and Restore; common mint/melt lifecycle rules have one shared reference.

`coco-browser-wallet` loads the general skill and adds IndexedDB, browser lifecycle, tab ownership, and optional React guidance. Future platform skills can reuse the shared wallet instructions. The README and Consumer Skills page document both skills and an explicit command to install the browser skill with its shared dependency.

Payment-method guidance covers reusable quote claiming, app-controlled partial claims, advertised onchain fee selection, and immediate or delayed settlement. Examples use public APIs and keep transaction ownership inside Coco.

Validation passed:
- Both skill structure validators, Prettier, and `git diff --check`.
- Skills CLI installation of both skills into a temporary project; standalone copies of the current skills have all dependency and reference links intact.
- All 16 TypeScript/TSX examples typechecked against the built workspace packages.
- 162 tests covering quote claimability and processing, BOLT12/onchain mint handlers, all three melt handlers, and the payment-request API.
- `bun run build` and `bun run docs:build`.

No changeset: this adds consumer skills and discovery docs without changing published package behavior or exports. Live-mint flows were not exercised.


igbopharaoh:
Great work here, I think a useful skill we can add can be a `coco expo wallet` focused skill for those building wallets with react native. Since the expo ecosystem is ingrained in the react native world and we support an ExpoSqulite db implementation, I think it'll be a good add.
Another one which can live under references can be a `multi-unit` skill and after I complete the work I have on it, `mintswap` skill and `batch minting` from your end .

Egge21M:
> Great work here, I think a useful skill we can add can be a `coco expo wallet` focused skill for those building wallets with react native. Since the expo ecosystem is ingrained in the react native world and we support an ExpoSqulite db implementation, I think it'll be a good add. Another one which can live under references can be a `multi-unit` skill and after I complete the work I have on it, `mintswap` skill and `batch minting` from your end .

Thank you! Yeah, the general idea is to have one base skill that explains all the general coco concepts and then to wrap that with platform specific skills. This PR introduces the base skill and the browser wrapper, but we should definitely include expo, node and bun too at some point. All new flows (e.g. MintSwap) should be added as a reference to the base skill and will then propagate automatically into the wrapper skills

igbopharaoh:
> Thank you! Yeah, the general idea is to have one base skill that explains all the general coco concepts and then to wrap that with platform specific skills. This PR introduces the base skill and the browser wrapper, but we should definitely include expo, node and bun too at some point. All new flows (e.g. MintSwap) should be added as a reference to the base skill and will then propagate automatically into the wrapper skills

Great, maybe in the long run, when someone works on new flows like mintswap, batch minting and co, a good to add after the implementation would be its skill. It's a bit more work but you've already created the template to make it easy to achieve

PR #495, opened 2026-09-14
refactor(storage): add nullable single-mint metadata lookup

### LINKED ISSUE #491: refactor(storage): add nullable single-mint metadata lookup
Follow-up from the design review of #489.

`StoredMintQueries.getMetadata()` and `RepositoryMintMetadataCommands.applyObservation()` currently call `getAllMints()` and then search for one mint URL. A lookup or refresh for one mint consequently loads every mint's metadata, including within the metadata transaction.

Evaluate and introduce a nullable single-mint lookup through the repository/query contracts, so absence can be represented without scanning all mints or catching arbitrary storage errors.

Suggested acceptance criteria:
- Metadata queries and scoped commands retrieve only the requested normalized mint URL.
- Missing mints return an explicit absent result; genuine storage failures propagate.
- The authoritative lookup in `applyObservation()` continues to use its transaction-scoped repository.
- Memory, SQL, and IndexedDB implementations provide equivalent behavior. Preserve existing throwing lookup callers or document any contract migration explicitly.
- Relevant conformance tests cover found/missing records and error propagation; assess public adapter type compatibility and changeset requirements.

This is a focused query improvement, not a request for an additional query wrapper layer or a merge blocker for #489.

bitsjack:
Closes #491

`StoredMintQueries.getMetadata()` and `RepositoryMintMetadataCommands.applyObservation()` called `getAllMints()` and scanned for one mint URL, loading every mint's metadata — including inside the metadata transaction — just to look up one.

Adds `MintRepository.findMintByUrl(mintUrl)`, returning the mint or `null`, implemented across all three backends (Memory, SQL, IndexedDB) by reusing each backend's existing single-mint lookup logic and returning `null` instead of throwing on a miss. The existing `getMintByUrl` is left unchanged since other callers rely on it throwing.

## Changes

- `packages/core/repositories/index.ts` — add `findMintByUrl` to the `MintRepository` interface
- `MemoryMintRepository`, `SqliteMintRepository` (sql-storage), `IdbMintRepository` (indexeddb) — implement it
- `StoredMintQueries.getMetadata()` and `RepositoryMintMetadataCommands.applyObservation()` — use it instead of `getAllMints().find(...)`
- Conformance tests for found/missing/deleted cases
- Changeset (`major` for `@cashu/coco-core` since it's a new required interface method for custom adapters, `patch` for sql-storage/indexeddb)

## Testing

- `bun run typecheck`: clean across all 9 packages
- `bun test packages/core/test/unit`: 1714 passed, 0 failed


packages/sql-storage/src/repositories/MintRepository.ts
```diff
@@ -41,6 +41,29 @@ export class SqliteMintRepository implements MintRepository {
     } satisfies Mint;
   }
 
+  async findMintByUrl(mintUrl: string): Promise<Mint | null> {
```
Egge21M, line 27: This duplicates `getMintByUrl` almost line by line. `getMintByUrl` should instead call `findMintByUrl` and throw on `null`

packages/indexeddb/src/repositories/MintRepository.ts
```diff
@@ -28,6 +28,21 @@ export class IdbMintRepository implements MintRepository {
     } satisfies Mint;
   }
 
+  async findMintByUrl(mintUrl: string): Promise<Mint | null> {
```
Egge21M, line 22: Same as https://github.com/cashubtc/coco/pull/495/changes#r4015347129

packages/core/repositories/index.ts
```diff
@@ -31,6 +31,8 @@ export interface ProofUnitFilter {
 export interface MintRepository {
   isTrustedMint(mintUrl: string): Promise<boolean>;
   getMintByUrl(mintUrl: string): Promise<Mint>;
+  /** Like {@link getMintByUrl}, but returns `null` instead of throwing when the mint is absent. */
```
Egge21M: The linked method does not have a doc string. Consider replacing the link with a standalone doc string (and add one to `getMintByUrl` if you like.

Egge21M (changes requested):
Thank you for taking this on. The general direction is exactly what we need. However there is an opportunity for reducing repetition that I outlined in the review. Also we are missing test coverage for the bundled storage adapters (memory is already covered)

packages/sql-storage/src/repositories/MintRepository.ts
bitsjack: Fixed in 2ea7499 — `getMintByUrl` now delegates to `findMintByUrl` and throws on `null`, instead of duplicating the query.

packages/indexeddb/src/repositories/MintRepository.ts
bitsjack: Fixed in 2ea7499 — same fix as sql-storage, `getMintByUrl` now delegates to `findMintByUrl` and throws on `null`.

packages/core/repositories/index.ts
bitsjack: Fixed in 2ea7499 — replaced the `{@link}` reference with a standalone doc string on `findMintByUrl`, and added one to `getMintByUrl` as well.

bitsjack:
Pushed 2ea7499 addressing the review:

- `SqliteMintRepository.getMintByUrl` and `IdbMintRepository.getMintByUrl` now delegate to `findMintByUrl` and throw on `null`, instead of duplicating the lookup query.
- Replaced the `{@link getMintByUrl}` doc reference with a standalone doc string on `findMintByUrl`, and added one to `getMintByUrl` too.
- Added conformance tests (found / missing / deleted) for the SQL and IndexedDB adapters, matching the existing memory-adapter coverage.

`bun run typecheck` is clean across all 9 packages and the relevant unit suites pass locally.

Egge21M (changes requested):
Thanks for addressing the duplication and documentation comments. The new IndexedDB tests aren’t picked up by CI: they use bun:test under src/repositories/, while IndexedDB CI runs Vitest with include: ['src/test/**/*.test.ts'].
Please move these cases into the existing browser contract suite in packages/indexeddb/src/test/contract.test.ts, reusing createRepositories() and createDummyMint(). That exercises real IndexedDB and runs through our existing Chromium, Firefox, and WebKit setup without CI changes.
Please also cover a storage-read failure to verify that genuine errors propagate rather than returning null. The browser verification command is documented in CONTRIBUTING.md.

bitsjack:
Pushed f5c6431 addressing the review:

- Moved the `findMintByUrl` IndexedDB tests from `src/repositories/MintRepository.test.ts` (bun:test, not picked up by CI) into `src/test/contract.test.ts`, reusing `createRepositories()`/`createDummyMint()` so they run through the real Chromium/Firefox/WebKit setup.
- Added a case forcing a Dexie `'reading'` hook to throw, verifying `findMintByUrl` propagates genuine storage errors instead of returning `null`.
- Added the equivalent storage-read-failure coverage to the SQL adapter (`sql-storage/src/test/MintRepository.test.ts`) for consistency, even though it wasn't flagged in this review — `findMintByUrl` has the same throw-vs-null contract there.

`bun run typecheck` is clean and the indexeddb contract suite / sql-storage suite pass locally.

Egge21M (approved):

